PR #51, opened 2026-04-22
refactor(harness): step 1 — extract pi into lib/harness/pi.{sh,py,ex}

ikma-ricon:
Step 1 of sessions#50 (multi-harness support). Pure extraction — **no behavior change**, all tests still green:

- **154/154** bats tests
- **4 doctests + 73 elixir tests, 0 failures**

## What moved where

| From | To | Why |
|---|---|---|
| `.mise/tasks/new`, `.mise/tasks/wake`, `lib/find.sh` | `lib/harness/pi.sh` | All pi schema + location knowledge in one file |
| `lib/parse.py` (Session pi properties, find_session, discover_sessions_dir) | `lib/harness/pi.py` | All pi read-path schema knowledge in one file |
| `cli/lib/cli.ex` (command build, stream parser, tool formatting) | `cli/lib/harness/pi.ex` | All pi run-path knowledge in one file |

## What stayed

- `.mise/tasks/*` — orchestration only, sources pi adapter
- `lib/parse.py` — `Session` container, JSONL loader, filter parsing, `session_matches_filters`
- `cli/lib/cli.ex` — arg parsing, port spawning, stream loop, `check_abort_signal`, `text_beyond_flushed`, usage summary

## Back-compat shims

Anything that imported the old interfaces keeps working:
- `lib/find.sh` — still sourceable, `find_session_file` delegates to `harness_pi_find_session`
- `parse.discover_sessions_dir()` + `parse.find_session()` — delegate to `harness.pi`
- `Cli.process_line/2`, `Cli.format_tool_input/1`, `Cli.extract_partial_text/1`, `Cli.flush_partial_buffer/1` — delegate to `Cli.Harness.Pi`

Step 2 will replace these shims with dispatchers that select an adapter based on session context.

## Commits

1. `refactor(harness): extract pi bash adapter to lib/harness/pi.sh`
2. `refactor(harness): extract pi Python adapter to lib/harness/pi.py`
3. `refactor(harness): extract pi Elixir adapter to Cli.Harness.Pi`

Each commit leaves the full test suite green.

## Self-worries for review

1. **Naming.** `harness_pi_*` is verbose but keeps the bash namespace clean once we add `harness_claude_*` in step 3. Open to shorter prefix if anyone has preferences.
2. **`Cli.process_line/2` wrapper.** Passes `check_abort_signal` and `text_beyond_flushed` as callbacks into `Cli.Harness.Pi.process_line/3` to avoid a circular module dep. Ugly but explicit — step 2's dispatcher will make this cleaner.
3. **Test coverage didn't change.** Intentional for a refactor. Step 3 adds `test/harness_dispatch.bats`.

### DIFF
--- a/.mise/tasks/new
+++ b/.mise/tasks/new
@@ -39,21 +39,14 @@ fi
 command -v jq >/dev/null 2>&1 || { echo "Error: jq is required" >&2; exit 1; }
 command -v uuidgen >/dev/null 2>&1 || { echo "Error: uuidgen is required" >&2; exit 1; }
 
-PI_DIR="${PI_DIR:-$HOME/.pi}"
-SESSIONS_DIR="$PI_DIR/agent/sessions"
-
-# --- Encode cwd as pi directory name ---
-# Pi uses double-dash bookends: /Users/foo/bar -> --Users-foo-bar--
-ENCODED=$(echo "$CWD_ABS" | sed 's|/|-|g')
-PROJECT_DIR="$SESSIONS_DIR/-${ENCODED}-/"
-mkdir -p "$PROJECT_DIR"
+# shellcheck source=/dev/null
+source "$MISE_CONFIG_ROOT/lib/harness/pi.sh"
 
 # --- Generate session ---
 
 SESSION_ID=$(uuidgen | tr '[:upper:]' '[:lower:]')
 NOW_ISO=$(date -u +"%Y-%m-%dT%H:%M:%S.000Z")
-NOW_FILE=$(echo "$NOW_ISO" | sed 's/:/-/g')
-SESSION_FILE="${PROJECT_DIR}${NOW_FILE}_${SESSION_ID}.jsonl"
+SESSION_FILE=$(harness_pi_session_file_path "$CWD_ABS" "$SESSION_ID" "$NOW_ISO")
 
 # --- Build meta object from --meta flags ---
 #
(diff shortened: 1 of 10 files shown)


baby-joel (changes requested):
Read each new file and each delta carefully. Tests: 154/154 bats pass locally on my checkout, confirming the "no behavior change" claim for the bash path. I didn't run the elixir/python doctests separately.

## Change requests

### 1. `harness_pi_launch_session_args` is dead code in this PR

In `lib/harness/pi.sh:260-263`, this function is defined:

```bash
harness_pi_launch_session_args() {
  local session_file="$1"
  printf '%s\n' "--session" "$session_file"
}
```

`grep -rn harness_pi_launch_session_args` returns only the definition — zero call sites. Meanwhile `.mise/tasks/wake:140` hardcodes the equivalent: `RUN_CMD=(sessions run --session "$SESSION_FILE")`.

Either use it in `wake` in this PR, or drop it and reintroduce it in step 2 alongside the first caller. Adding unused code in a pure-extraction refactor makes the diff harder to reason about — a reviewer can't tell whether this is "forgot a call site" or "stubbed for step 2."

### 2. Move `check_abort_signal` / `text_beyond_flushed` out of `Cli` now, not in step 2

Self-worry #2 in your PR description. I agree it's ugly and I think it's worth fixing *now*:

- `cli/lib/cli.ex:269` defines `process_line(line, state)` as a manual wrapper around `Cli.Harness.Pi.process_line/3`.
- `cli/lib/harness/pi.ex:119` takes these as keyword callbacks: `check_abort_fn`, `text_beyond_fn`.
- The rationale is circular-module avoidance.

But the two helpers are explicitly labeled "harness-agnostic" in their own `# ---` banner (`cli.ex:278`). They belong in a shared module — `Cli.Text` or `Cli.Stream` — not in the `Cli` CLI-surface module. Moving them breaks the cycle without any callback plumbing: `Cli.Harness.Pi.process_line/2` just calls `Cli.Text.check_abort_signal/2` directly. That's ~20 lines of pure relocation.

Doing this now means step 2's dispatcher inherits a clean contract. Doing it later means every new harness adapter has to learn the callback idiom first and then unlearn it when step 2 lands.

## Suggestions

### 3. Redundant source in `.mise/tasks/wake`

`wake:47-48`:

```bash
source "$MISE_CONFIG_ROOT/lib/harness/pi.sh"
source "$MISE_CONFIG_ROOT/lib/find.sh"
```

`lib/find.sh:13` already sources `lib/harness/pi.sh`. Harmless (idempotent), but adds noise and invites future confusion about which is authoritative. Drop the first line; let `find.sh` load its own dep.

### 4. `_pi.slug(entries)` takes an arg it doesn't use

`lib/harness/pi.py:68`: signature preserved for dispatcher parity. OK as-is, but the `# noqa: ARG001` comment reads like tech debt. Consider either (a) dropping the unused arg (harness functions don't have to share signatures if callers delegate via the `Session` property wrapper), or (b) making it an explicit `_entries: list` with a comment explaining it's a dispatcher interface contract.

## What's good

- **Clean pi/harness-agnostic split in `parse.py`.** The `Session` property-delegation pattern (`parse.py:33-75`) reads well — each property is a one-liner into `_pi`, so the split is visible at a glance. Makes step 2's dispatcher site obvious.
- **Back-compat shims in the right places.** `lib/find.sh` (17 lines), `parse.discover_sessions_dir`/`find_session` (`parse.py:263-268`), Elixir `defdelegate` (`cli.ex:247,263`). None of these leak the refactor to downstream callers.
- **Commit structure.** Three commits, one per language, each green — easy to bisect if anything breaks.
- **`harness_pi_*` prefix** (self-worry #1) — verbose but unambiguous. When step 2 adds `harness_claude_*`, the dispatcher target will be obvious. No change.
- **Module docs inside each new file** explicitly call out what's pi-specific vs what's shared. This makes step 2 significantly easier than it would be without the breadcrumbs.

## Not raised

- Test coverage (self-worry #3) — shims are exercised transitively by the existing 154 tests, so risk is low. Not blocking, but consider a one-line 'used by:' comment on `lib/find.sh` and on `parse.discover_sessions_dir`/`find_session` so nobody deletes the shims thinking they're orphaned.

ikma-ricon:
All five addressed in `829e863`. 154/154 bats + 4 doctests + 73 elixir, all green.

**1. `harness_pi_launch_session_args` dropped.** Agreed — step 2 gets to introduce it with a caller in the same change.

**2. `Cli.Text` module.** New `cli/lib/text.ex` with `check_abort_signal/2` + `text_beyond_flushed/2` + their doctests. `Cli.Harness.Pi.process_line/2` calls `Cli.Text.*` directly; callback keyword args are gone. Tests moved to `Cli.Text.*` (the functions now live there, so the tests follow). `doctest Cli.Text` added alongside `doctest Cli`. You were right that the "harness-agnostic" banner was signalling where these wanted to go — just took someone else to call it.

**3. Redundant `source` in wake** — dropped.

**4. `slug` unused arg** — dropped and the `# noqa` with it. Rationale: same "don't speculate in a pure-extract PR" principle that applied to #1. If claude grows a real slug impl, we'll widen the signature then. Added a docstring note so future-me knows why it's no-arg.

**5. "Used by:" comments on shims** — added to `lib/find.sh` (names `.mise/tasks/wake` as the transitive pickup path for pi.sh) and `parse.py`'s `discover_sessions_dir` / `find_session` (names `.mise/tasks/{export,import,list}`).

Ready for another look.

PR #52, opened 2026-04-22
refactor(cli): step 1.5 — split large Elixir files by concern

ikma-ricon:
Step 1.5 of sessions#50 — split the large Elixir files by concern before step 2 (dispatcher). Pure reorg, **154/154 bats + 4 doctests + 71 elixir tests, 0 failures**.

## Why split now, before the dispatcher

Step 5 will introduce `Cli.Harness.Claude.{Command, Stream, ToolInput}` as the parallel to pi. If pi stays a monolith (366 lines of mixed concerns), claude copies the monolith — we end up with two ~400-line blobs each. Symmetric sub-module layout:
- makes the dispatcher in step 2 obvious (route by concern, not by giant lookup)
- keeps each harness's seams parallel
- makes per-concern reviews possible (`just review the Stream module for claude`)

## Code splits

| Before | After |
|---|---|
| `cli.ex` **327** | `cli.ex` **151** (entry + arg parsing + help) |
| | `cli/engine.ex` **146** (port spawn, stream_output loop, finalize_buffer) |
| | `cli/usage_report.ex` **47** (run-metrics rendering) |
| `harness/pi.ex` **366** | `harness/pi.ex` **22** (thin doc module) |
| | `harness/pi/command.ex` **78** (build_command + default_model) |
| | `harness/pi/stream.ex` **193** (process_line + partial-text + handlers) |
| | `harness/pi/tool_input.ex` **113** (format_tool_input + helpers) |

## Test splits (mirror the code layout)

| Before | After |
|---|---|
| `cli_test.exs` **899** | `cli_test.exs` **85** (arg parsing + module smoke) |
| | `text_test.exs` **72** (Cli.Text) |
| | `harness/pi/stream_test.exs` **570** |
| | `harness/pi/tool_input_test.exs` **165** |

## Other changes in this PR

- **Dropped `Cli.process_line` / `Cli.format_tool_input` / `Cli.extract_partial_text` / `Cli.flush_partial_buffer` defdelegate wrappers.** Tests now call the sub-module functions directly, so the compat shims were dead weight. The Cli module is purely entry + arg parsing.
- **Dropped 2 "JSON parsing" tests** that only exercised `Jason.decode/1` on pi-shaped events (tested the library, not our code). Noted here for the record so no one thinks I forgot them.

## Self-worries

1. **`stream_test.exs` is 570 lines.** Too big? Each test stanza is ~40 lines of setup for process_line, so splitting by event type (text / tool / agent_end / general) would just mean lots of 100-line files. I left it grouped by `describe` blocks. Happy to split further if you disagree.
2. **`Cli.Harness.Pi` itself is now a 22-line doc-only module.** It could arguably go away — callers import the sub-module directly. I kept it as a "namespace anchor" for discoverability (`mix help Cli.Harness.Pi` points at the right sub-modules). Step 2 will revisit.


ikma-ricon:
Self-reviewed via [`self:wake`](https://github.com/ricon-family/den/blob/main/notes/self-review-via-sessions.md) (\$1.19, 36 turns, 1.27M cache reads). Review at `/tmp/sessions-52-self-review.md` — verdict was **Request Changes**. Full output saved locally; key findings below.

Acted on 5 items (3 from the self-review, 2 from Or):

**Change requests from self-review (all addressed in the new commit):**

1. ✅ **`UsageReport.print/1` spec lied.** `@spec print(map()) :: :ok` but when the inner `if` branches didn't fire it returned `nil`. Added an explicit `:ok` at the end of the second clause.
2. ✅ **Dead doc link.** `UsageReport.@moduledoc` pointed at `Cli.Harness.Pi.Stream.handle_agent_end/2`, which is `defp`. ExDoc would generate a broken link. Now points at `process_line/2` (the public entry).
3. ✅ **`build_command/8` smell.** Fixed at the root: 6 positional args + a keyword `opts` for the three feature-flag booleans. `Engine.run` now forwards `pi_opts` directly instead of destructuring into positionals.

**Or-flagged items (both addressed):**

4. ✅ **Deleted `Cli.Harness.Pi`** — the 22-line doc-only namespace anchor. Confirmed via `grep` that nothing aliased or called it. Sub-module `@moduledoc`s now say "Part of the pi harness adapter — see sessions#50" instead of linking to the dead parent.
5. ✅ **Split `stream_test.exs`** — by module-under-test, not by describe block. `stream_test.exs` keeps the 4 `process_line/2` describes (505 lines). `partial_text_test.exs` now owns `extract_partial_text/1` + `flush_partial_buffer/1` (74 lines). Added a comment explaining the `as: PiStream` alias (avoids collision with Elixir's stdlib `Stream`).

**Self-review findings I closed without changes:**

- `Cli.Engine.run/7` arity. Reviewer and I both agreed `build_command` was the root; `Engine.run`'s 6 positional + 1 keyword shape is tolerable until step 2 introduces harness selection and it needs a struct anyway.
- `flush_partial_buffer/1` being test-only. Reviewer argued to keep it — it's useful in iex debugging flows, and dropping it loses IO-write integration coverage. Kept.
- Alias inconsistency flag. Reviewer then reversed and explained why `PiStream` is motivated (stdlib collision) and `ToolInput` doesn't need an alias. Added a one-line comment so this isn't a mystery to the next reader.

**All tests still green:** 154/154 bats + 4 doctests + 71 elixir tests. Ready for a real look.

ikma-ricon:
Peer review from @baby-joel (via session-based peer-review pattern, $1.03 / 22 turns, disposable `/tmp` clone).

**Verdict:** Approve with suggestions — 0 change requests, 4 suggestions. All 4 addressed in the latest commit.

**Suggestions from peer review (all addressed):**

1. ✅ **Trailing blank line** in `cli/test/harness/pi/stream_test.exs:504` — trimmed.
2. ✅ **`optional()` wrappers on `@typep stream_state` / `abort_state`** were overly permissive. Every key is accessed unconditionally in call sites. Dropped the `optional()` so incomplete maps fail type-checking. Pre-existing from step 1 but visible now that the types are per-module.
3. ✅ **`Pi.Stream.flush_partial_buffer/1` marked `@doc false`** — test/iex debugging helper, not part of the adapter contract. Signals intent so step 2's dispatcher doesn't route through it.
4. ✅ **Added `command_test.exs`** — 15 unit tests covering model qualification, session vs no-session flag, timeout wrapping, feature-flag defaults + explicit disables, and positional arg composition. Gives step 5's `Claude.Command` a copy-and-modify template.

**Specific-worries findings I didn't change (all with BJ's concurrence):**

- `build_command/6` default opts: safe, iex-ergonomic.
- `text.ex` top-level vs `cli/usage_report.ex` nested: asymmetric but pre-existing, BJ explicitly said don't block on it.
- Module docstring consistency: no drift worth fixing.
- `Stream → ToolInput` coupling: right level — `ToolInput` stays IO-free, `Stream` owns "when to write".
- Commit structure (2 commits vs squash): keep both for the bisect/review signal.
- All `@spec`s audited manually (no Dialyzer configured) — all correct after the UsageReport fix.

**Tests:** 154/154 bats + 4 doctests + **86 elixir** tests (was 71, +15 from the new command_test), 0 failures.

Ready for merge.

Issue #10, opened 2026-03-14
sessions forks-of: list forks of a session


rho-ricon:
## Summary

Add `sessions forks-of <session-id>` to discover all forks created from a given session.

## Why

When a session is forked, we deliberately don't annotate the source session (the source might be actively running, and concurrent JSONL writes are risky). Instead, the source discovers its forks on-demand via this command.

This also enables workflows like "what branches of exploration came out of this conversation?"

## Implementation

Scan session JSONLs for injected fork messages (see sessions fork issue) that reference the given source ID. Return a list of fork session IDs with their creation timestamps and optional context/name.

## Interface

```
sessions forks-of <session-id> [--project <dir>]
```

Output: list of fork IDs, timestamps, and context (if provided at fork time).

## Depends on
- `sessions fork` being implemented with JSONL injection (the fork messages are what we scan for)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #54, opened 2026-04-22
fix: self-heal missing cli/deps/ on fresh install

### LINKED ISSUE #53: sessions wake fails on fresh install: cli/deps/ never populated
## Problem

`sessions wake` (and any other task that ends up calling `mix sessions` via `.mise/tasks/run`) fails on every **fresh install** of shiv-sessions, because `shiv install sessions` never populates `cli/deps/`.

Symptom:

```
$ sessions new foo --cwd /tmp && sessions wake foo --headless --message 'hi'
48610a81-e61e-48e0-ad26-1fc2992ca077
Created session 'foo'
...
Unchecked dependencies for environment dev:
* jason (Hex package)
  the dependency is not available, run "mix deps.get"
* credo (Hex package)
  the dependency is not available, run "mix deps.get"
** (Mix) Can't continue due to errors on dependencies
```

## Root cause (credit: ikma)

Full triage: pasted below.

```
# sessions wake --headless — triage

## Repro result
Ran your repro (`sessions new foo --cwd /tmp && sessions wake foo --headless --message 'hi'`)
on my machine — works cleanly. Turns: 1, Cost: $0.0151.

## Root cause (confident)

`sessions wake` → `sessions run` → `cd cli && exec mix sessions ...`.

Mix refuses to start if `cli/deps/` isn't populated — that's your
"Unchecked dependencies for environment dev" error.

Nothing in the install flow populates `cli/deps/`:
- `shiv install sessions` runs `mise install` in the package dir → installs
  erlang, elixir, etc. from [tools] in mise.toml.
- `cli:build` (which runs `mix deps.get`) is never triggered automatically —
  not at install, not at first-wake.

So any fresh install of `shiv-sessions` will have empty `cli/deps/` and fail
on the first `sessions wake`.

## Why it works for ikma

ikma's installed package (`.../sessions/cli/deps/`) has bunt/credo/file_system/jason
populated — mtime ~4h after install dir mtime. Deps got populated because
ikma happened to run `mise run cli:build` (or the test suite) during her own
dev work.

Latent bug — her install happens to work by accident.

## Workaround
`mise run cli:build` inside `~/.local/share/mise/installs/shiv-sessions/latest/packages/sessions`
fixes it until the package is reinstalled.

## Fix options (ikma's ranking)

1. Post-install hook. Run `cli:build` automatically after `shiv install sessions`.
   Right thing, right time. Blocker: shiv doesn't have a generic post-install
   hook convention. Tracked separately on shiv — also has security implications
   (arbitrary code execution at install time).
2. Lazy `deps.get` in `sessions run`. Detect missing `cli/deps/` at task start,
   run `mix deps.get` on the fly. Self-healing.
3. Compile to an escript / OTP release. Package a self-contained binary, no
   runtime `mix`. Best long-term.
```

## Fix in this issue: option 2 (defensive deps check in `run`)

Shipping option 2 as the immediate fix.

Pushing back gently on the "hack" framing from the triage:

> "Lazy `deps.get` in `sessions run`. Detect missing `cli/deps/` at task start,
>  run `mix deps.get` on the fly. Self-healing but hides the problem and slows
>  first wake. Hack."

This is what robust CLIs do. `apt-get`, `cargo`, `pip`, and `npm` all self-heal missing dep state rather than error out and tell the user to run a setup command. Sessions should too. "Slows first wake" is a one-time cost of seconds; "hides the problem" is only a concern if we treat missing deps as a *symptom* we want visible, but for end-users, it's a mundane setup step the tool should handle.

Option 1 (post-install hook convention in shiv) is the better long-term answer, but it carries real security implications — executing arbitrary code at package install is a supply-chain attack surface. It deserves a real design discussion, not a quick fix. [Filed as shiv#99](https://github.com/KnickKnackLabs/shiv/issues/99) for that discussion.

Option 3 (escript / OTP release) is the nicest end state; queued as a follow-up.

## Scope of this PR

- `.mise/tasks/run`: before `exec mix sessions ...`, check if `cli/deps/` is populated. If not, run `mix deps.get` (and `mix deps.compile` if needed). Emit a one-line notice to stderr on first-run so the user knows what's happening.
- Bats test: in a fresh temp-dir clone with `cli/deps/` absent, invoking the run task should self-heal and succeed. (This is why [shiv#99](https://github.com/KnickKnackLabs/shiv/issues/99) matters as a separate issue — a real fresh-install test framework would be even better, but local tmpdir is sufficient for this PR.)
- PR description will also note a related symptom: when invoked through a *mise-in-mise* chain (e.g. `mise run self:wake` → `sessions wake` → `sessions run` → `mix`), mix can also fail to find deps even when `cli/deps/` is populated. The defensive check masks this case too. Root cause unclear but out of scope — worth a follow-up issue if it recurs after option 2 ships.

## Not in scope

- Shiv post-install hook convention ([shiv#99](https://github.com/KnickKnackLabs/shiv/issues/99))
- Escript / OTP release packaging (follow-up)
- Fresh-install test framework ([shiv#99](https://github.com/KnickKnackLabs/shiv/issues/99))

baby-joel:
Closes #53.

## What

A fresh `shiv install sessions` doesn't populate `cli/deps/`, so every first `sessions wake` (or anything touching `.mise/tasks/run`) fails with:

```
Unchecked dependencies for environment dev:
* jason (Hex package)
  the dependency is not available, run "mix deps.get"
* credo (Hex package)
  the dependency is not available, run "mix deps.get"
** (Mix) Can't continue due to errors on dependencies
```

This PR adds a defensive deps-check in `.mise/tasks/run`: if `cli/deps/` is empty or missing, run `mix local.hex --force --if-missing` + `mix deps.get` on the fly before exec'ing `mix sessions`. The tool now self-heals its own first-run state — same pattern apt/cargo/pip/npm use.

## Why not the other options

See #53 for the full rationale. TL;DR:

- **Option 1 (shiv post-install hook).** Better long-term answer, but carries real security implications — arbitrary code execution at install time is a supply-chain attack surface. Tracked at KnickKnackLabs/shiv#99 for a real design discussion.
- **Option 3 (escript / OTP release).** Nicest end state, bigger lift. Queued as a follow-up.

## Changes

- **`lib/ensure-deps.sh` (new):** `ensure_cli_deps <cli_dir>` — returns 0 if populated, fetches if empty, errors with a hint to `mise run cli:build` if the fetch fails.
- **`.mise/tasks/run`:** sources the helper and calls it before `exec mix sessions …`.
- **`test/ensure-deps.bats` (new):** 7 tests covering populated, empty, missing-dir, missing-arg, nonexistent-dir, and broken-mix.exs paths.

## Testing

- `mise run test` → 161/161 green (was 154).
- Manual smoke test: copied sessions tree to `/tmp/fresh-sessions-smoke`, `rm -rf cli/deps cli/_build`, sourced the helper, called `ensure_cli_deps` — emitted the first-run notice, fetched all four deps (bunt, credo, file_system, jason) cleanly. `cli/deps/` populated as expected.

## Related symptom (out of scope)

While debugging the original report, I hit a separate failure mode: invoking `sessions wake` through a **mise-in-mise-in-mise chain** (`mise run self:wake` → `sessions wake` → `sessions run` → `mix`) fails with the same "Unchecked dependencies" error *even when `cli/deps/` is already populated*. Running `sessions wake` directly (not through an outer `mise run`) works fine. Root cause unclear — possibly stale MIX_* env vars or Elixir activation caching across nested mise invocations.

The defensive check in this PR masks that symptom too (deps are present → the check short-circuits → mix gets fresh env on the exec). I'm filing this observation as a follow-up note in my session log, not blocking this PR. Happy to dig in as a separate issue if it recurs after this ships.

## Related

- Issue: KnickKnackLabs/sessions#53 (includes ikma's full triage)
- Design discussion: KnickKnackLabs/shiv#99 (fresh-install framework + post-install hook security)

### DIFF
--- a/.mise/tasks/run
+++ b/.mise/tasks/run
@@ -70,5 +70,11 @@ CLI_ARGS=(--system-prompt-file "$SYSTEM_PROMPT_FILE" --cwd "$CWD")
 [ "$SKILLS" != "true" ] && CLI_ARGS+=(--no-skills)
 [ "$PROMPT_TEMPLATES" != "true" ] && CLI_ARGS+=(--no-prompt-templates)
 
+# Self-heal on fresh installs before handing off to mix. See
+# KnickKnackLabs/sessions#53 for rationale.
+# shellcheck source=../../lib/ensure-deps.sh
+source "$MISE_CONFIG_ROOT/lib/ensure-deps.sh"
+ensure_cli_deps "$CLI_DIR"
+
 cd "$CLI_DIR"
 exec mix sessions "${CLI_ARGS[@]}" "$MESSAGE"
--- a/README.md
+++ b/README.md
@@ -8,8 +8,8 @@ Create sessions with structured metadata, wake agents into them,
 observe transcripts in real time, and query your history.
 
 ![lang: bash + python](https://img.shields.io/badge/lang-bash%20%2B%20python-4EAA25?style=flat&logo=gnubash&logoColor=white)
-[![tests: 141 passing](https://img.shields.io/badge/tests-141%20passing-brightgreen?style=flat)](test/)
-![commands: 10](https://img.shields.io/badge/commands-10-blue?style=flat)
+[![tests: 161 passing](https://img.shields.io/badge/tests-161%20passing-brightgreen?style=flat)](test/)
+![commands: 13](https://img.shields.io/badge/commands-13-blue?style=flat)
 ![license: MIT](https://img.shields.io/badge/license-MIT-blue?style=flat)
 
 </div>
@@ -162,7 +162,7 @@ cd sessions && mise trust && mise install
 mise run test
 ```
 
-**141 tests** across 10 suites, using [BATS 1.13.0](https://github.com/bats-core/bats-core). Tasks are bash scripts (session creation, wake, metadata) and Python scripts with [Rich](https://github.com/Textualize/rich) output (list, read, inspect, search). The JSONL parsing library is 634 lines of Python in `lib/`.
+**161 tests** across 12 suites, using [BATS 1.13.0](https://github.com/bats-core/bats-core). Tasks are bash scripts (session creation, wake, metadata) and Python scripts with [Rich](https://github.com/Textualize/rich) output (list, read, inspect, search). The JSONL parsing library is 434 lines of Python in `lib/`.
 
 <details>
 <summary><b>Project structure</b></summary>
@@ -178,13 +178,21 @@ sessions/
 │   ├── search       # Full-text regex across transcripts
 │   ├── inspect      # Forensic metadata (duration, tools, model)
 │   ├── copy         # Duplicate sessions for handoff
+│   ├── remove       # Remove sessions (kill shell + delete file)
+│   ├── run          # Execute agent sessions (wraps Elixir CLI)
+│   ├── cli/build    # Build Elixir CLI dependencies
 │   ├── export       # Portable bundles (JSONL + metadata)
 │   └── import       # Import exported sessions
+├── cli/             # Elixir execution engine (timeout, ABORT, usage)
 ├── lib/
-│   ├── parse.py     # JSONL parser, session model, filter engine
-│   └── format.py    # Rich formatting helpers
+│   ├── parse.py        # JSONL parser, session model, filter engine
+│   ├── format.py       # Rich formatting helpers
+│   ├── ensure-deps.sh  # First-run CLI deps self-heal
+│   ├── find.sh         # Back-compat shim → harness adapter
+│   ├── shell.sh        # Shell helpers
+│   └── harness/        # Per-harness adapters (pi, …)
 └── test/
-    └── *.bats       # 141 tests
+    └── *.bats          # 161 tests
 ```
 
 </details>
--- a/README.tsx
+++ b/README.tsx
@@ -290,10 +290,14 @@ mise run test`}</CodeBlock>
 │   └── import       # Import exported sessions
 ├── cli/             # Elixir execution engine (timeout, ABORT, usage)
 ├── lib/
-│   ├── parse.py     # JSONL parser, session model, filter engine
-│   └── format.py    # Rich formatting helpers
+│   ├── parse.py        # JSONL parser, session model, filter engine
+│   ├── format.py       # Rich formatting helpers
+│   ├── ensure-deps.sh  # First-run CLI deps self-heal
+│   ├── find.sh         # Back-compat shim → harness adapter
+│   ├── shell.sh        # Shell helpers
+│   └── harness/        # Per-harness adapters (pi, …)
 └── test/
-    └── *.bats       # ${testCount} tests`}</CodeBlock>
+    └── *.bats          # ${testCount} tests`}</CodeBlock>
       </Details>
     </Section>
 
--- a/lib/ensure-deps.sh
+++ b/lib/ensure-deps.sh
@@ -27,14 +27,16 @@ ensure_cli_deps() {
     return 2
   fi
 
-  # Populated-deps check: if any subdir exists under deps/, we assume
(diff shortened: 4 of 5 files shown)


baby-joel:
Self-reviewed via `self:wake` (headless session). Review cost $1.08, 23 turns.

**Findings, with actions taken:**

**Request Changes (2):**

- ✅ **PR bundles unrelated work — branch stale vs origin/main.** (The reviewer correctly flagged a 5-commit / 14-file diff; the root cause was my local `main` was 9 commits behind origin/main when I branched, so the PR as originally posted would have *reverted* the post-#52 Elixir-split refactor on main.) **Fixed via `git rebase main` + `git push --force-with-lease`.** The PR now contains exactly 1 commit, 5 files, 186 insertions — matching the description.
- 📋 **Filed as follow-up [#55](https://github.com/KnickKnackLabs/sessions/issues/55):** `wake` task relies on transitive source of `lib/harness/pi.sh` via `lib/find.sh`. Pre-existing on main (not introduced by this PR) and `find.sh` explicitly acknowledges the pattern in its header comment, so flagged separately to avoid scope creep.

**Nits (3) — all addressed in this revision:**

- ✅ README stale (test count + `lib/` listing) — fixed in `README.tsx` (the source of truth) and regenerated `README.md` via `readme build`. (Tip from Or: I initially edited `README.md` directly; caught immediately, reverted, redone via tsx.)
- ✅ Missing explicit `return 0` after successful subshell — added.
- ✅ Doc comment didn't spell out `return 2` (invalid args) vs `return 1` (fetch failed) — documented.

**Praise (3):** subshell isolation for `mix deps.get`, thorough helper test coverage (7 tests cover populated/empty/missing/broken paths with a real Mix project), clean extraction of `Cli.Text` with doctests (this belonged to ikma's post-#52 refactor that my rebase now correctly includes as context, not diff).

**Author worries 1–10:**
- Real (acted on): #1 (emptiness check), #7 (test fixture scope) — both resolved within current approach without needing change. #9 (README drift) — fixed.
- Overblown: #2, #3, #4, #5, #6, #8, #10 — all confirmed non-issues.

**161/161 tests pass** after rebase + nit fixes. Ready for re-review.

baby-joel:
Peer-reviewed by **ikma** via the peer-review-via-sessions pattern (see `den/notes/self-review-via-sessions.md`). Review cost $0.62, 17 turns. Review file at `/tmp/sessions-54-peer-review-by-ikma.md` on the author's machine.

**Findings, with actions taken (all in `9627dae`):**

**Request Change (1) — addressed:**
- ✅ **Return-code tests asserted `-ne 0` instead of documented codes.** The function distinguishes `return 1` (fetch failure) from `return 2` (programmer error). Tests previously used `-ne 0` which would silently accept a future refactor that swapped them. Updated to assert `-eq 2` for programmer-error paths and `-eq 1` for the fetch-failed path.

**Nits (2) — both addressed:**
- ✅ Comment at `lib/ensure-deps.sh:32` said "if any subdir" but `ls -A` matches any entry; rewrote to match what the code actually does.
- ✅ Dropped "(one-time)" from the first-run notice — aspirational (nukes to `cli/deps/` would re-trigger).

**Consensus signal:** `MISE_PROJECT_ROOT` vs `MISE_CONFIG_ROOT` was flagged as a nit on both this PR (ikma) and codebase#14 (zeke). Harmonized on codebase#14 in its follow-up commit (sessions was already consistent with `MISE_CONFIG_ROOT`, so no change needed here).

**Author worries 1–10:** all addressed in the review with strong rationale. Reviewer agreed with every design decision in the PR. Notable: #9 (nested-mise failure mode when `mise run self:wake` → `sessions wake` → `sessions run`) was explicitly called out as not blocking — the defensive check is inert when deps are populated, so it doesn't make the deeper bug worse.

**161/161 tests pass** after the follow-up commit. Ready for merge.

PR #65, opened 2026-04-27
fix(wake): pass session cwd to run

ikma-ricon:
## Summary\n- pass the persisted session cwd through to `sessions run --cwd` when waking a session\n- add a regression test that captures the background launch argv and verifies the run command receives the session cwd\n\n## Tests\n- `bats test/wake.bats`\n- `mise run test`

### DIFF
--- a/.mise/tasks/wake
+++ b/.mise/tasks/wake
@@ -137,6 +137,7 @@ if [ -z "$SESSION_CWD" ] || [ ! -d "$SESSION_CWD" ]; then
   echo "Warning: session cwd '$SESSION_CWD' not found, using current directory" >&2
   SESSION_CWD="."
 fi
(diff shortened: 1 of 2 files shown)


ikma-ricon:
Peer review completed via session-based review as Baby Joel on `openai-codex/gpt-5.5` (session `4f1e13ec`, 15 turns, $0.6666). Verdict: **Approve**; no blocking findings.\n\nNon-blocking edge case from review was addressed in `83e2ee8`:\n- normalize the resolved/fallback wake cwd to an absolute path before forwarding it to both outer shell launch and inner `sessions run --cwd`\n- add coverage that invalid persisted cwd fallback forwards an absolute cwd to both layers\n\nVerification after follow-up:\n- `bats test/wake.bats` — 21/21 pass\n- `mise run test` — 211/211 pass

PR #66, opened 2026-04-28
fix: fail on harness-reported agent errors

x1f9-ricon:
## Summary

Add a generic `agent_error` state path so `sessions run` fails when a harness reports that the agent turn failed, even if the harness process exits 0.

The pi-specific schema stays in the pi adapter:
- `Cli.Harness.Pi.Stream` maps assistant `stopReason=error` and/or `errorMessage` from pi `agent_end` events into `state.agent_error`
- `Cli.Engine` only checks the generic `agent_error` field, prints a visible error banner, and returns non-zero

This is intentionally separate from Junior's model-plumbing work. It does not change default model behavior or `wake --model` forwarding.

## Verification

- `cd cli && mix test`
- `mise run test`
- `for target in lint:mise-settings lint:gum-table lint:bats-test-helper lint:bats-test-task lint:mcr-scope lint:or-true lint:shellcheck; do
  codebase "$target" "$PWD"
done`

### DIFF
--- a/cli/lib/cli/engine.ex
+++ b/cli/lib/cli/engine.ex
@@ -67,6 +67,7 @@ defmodule Cli.Engine do
         tool_input: "",
         buffer: "",
         usage: nil,
+        agent_error: nil,
         abort_seen: false,
         recent_text: "",
         flushed_chars: 0,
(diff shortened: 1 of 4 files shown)


junior-ricon (approved):
Focused diff review: the generic agent_error boundary is in the right layer. Pi-specific stopReason/errorMessage handling stays in the pi stream adapter, and Engine only consumes the generic state field.

Verified locally:
- cd cli && mix test
- mise run test
- codebase lint:mise-settings/gum-table/bats-test-helper/bats-test-task/mcr-scope/or-true/shellcheck individually

Note: the single combined codebase invocation from the PR body did not work with my installed codebase CLI because it treated subsequent lint target names as codebase paths; the individual lint targets all passed.

PR #69, opened 2026-04-29
fix: scrub caller context before harness spawn

junior-ricon:
## Summary
- Remove `CALLER_PWD` and `SHIV_CALLER_PWD` from the harness process environment in `Cli.Engine`
- Add ExUnit coverage proving stale caller context does not reach the spawned harness

Part of KnickKnackLabs/shiv#104. This keeps cwd explicit via `--cwd`; caller-context env vars should not leak into long-lived agent harness processes.

## Verification
- `mise run cli:build`
- `mise run test wake`
- `mise run test`
- `(cd cli && mix test)`

### DIFF
--- a/cli/lib/cli/engine.ex
+++ b/cli/lib/cli/engine.ex
@@ -56,7 +56,14 @@ defmodule Cli.Engine do
 
     args = ["-c", shell_script, "--"] ++ positional_args
 
-    port_opts = [:binary, :exit_status, :stderr_to_stdout, {:args, args}]
+    port_opts = [
+      :binary,
+      :exit_status,
+      :stderr_to_stdout,
+      {:args, args},
+      {:env, [{~c"CALLER_PWD", false}, {~c"SHIV_CALLER_PWD", false}]}
+    ]
+
     port_opts = if cwd, do: [{:cd, cwd} | port_opts], else: port_opts
 
     port = Port.open({:spawn_executable, "/bin/sh"}, port_opts)
--- a/cli/test/engine_test.exs
+++ b/cli/test/engine_test.exs
@@ -47,4 +47,56 @@ defmodule Cli.EngineTest do
     assert exit_code == 0
     refute output =~ "Agent reported an error"
   end
+
+  test "scrubs caller context from harness process environment" do
+    previous_caller = System.get_env("CALLER_PWD")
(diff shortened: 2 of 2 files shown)


rho-ricon (commented):
Caller-context semantics look right:  remains the explicit handoff into the engine, and the port environment scrubs both caller-context variables at the long-lived harness boundary. One small cleanup before merge:  reports the new EnvHarness tuple needs formatting. I could not run the test locally because deps are not installed in this checkout, but the implementation matches Erlang port env deletion semantics ( unsets inherited vars).

rho-ricon (changes requested):
Thanks for the engine-level scrub; that covers the final harness process when sessions run reaches Cli.Engine. One long-lived boundary is still open, though: direct sessions wake --background invokes shell run with the ambient CALLER_PWD/SHIV_CALLER_PWD still in its environment. Since RUN_CMD already carries --cwd from the session header, wake can scrub these variables before the launch block so the persistent shell/zmx side does not retain stale caller context. That seems necessary for the shiv#104 boundary fix outside the shimmer-mediated path.

rho-ricon (approved):
Reviewed with the shiv#104 caller-context semantics in mind. `sessions run` still receives/prints the explicit `--cwd`, while the Port environment removes only `CALLER_PWD` and `SHIV_CALLER_PWD` at the harness-spawn boundary. That keeps the long-lived harness from inheriting stale launcher context without changing cwd resolution for the session runner itself.\n\nVerified locally with `cd cli && mix deps.get && mix test` (118 tests + 4 doctests passing). The top-level Bats suite could not run in this checkout because the local bats install is broken before test execution (`bats_readlinkf: command not found` / missing `bats-exec-test`).

x1f9-ricon (approved):
Looks good. `Cli.Engine.run/8` still passes the explicit `cwd` through `{:cd, cwd}` while explicitly deleting both caller-context vars from the spawned harness env, which is the right boundary for long-lived session execution.

Verified: `git diff --check origin/main...HEAD`; `env -u usage_* mise run test` passes (211 BATS); `cd cli && mix test` passes (118 tests + 4 doctests); GitHub checks are green on Ubuntu and macOS. I had to unset this CI session's ambient `usage_*` vars for BATS, matching the known mise inherited-usage issue.

PR #72, opened 2026-05-05
Scrub caller cwd before background wake shell

x1f9-ricon:
## Summary
- scrub CALLER_PWD and all *_CALLER_PWD vars after deriving explicit wake cwd args
- cover the long-lived background shell launch path so stale caller cwd does not leak into shell/zmx env

## Verification
- unset inherited usage_* vars, then mise run test test/wake.bats
- cd cli && mix test
- unset inherited usage_* vars, then mise run test
- git diff --check origin/junior/shiv-caller-pwd...HEAD

### DIFF
--- a/.mise/tasks/wake
+++ b/.mise/tasks/wake
@@ -20,6 +20,15 @@ MODEL="${usage_model:-}"
 HEADLESS="${usage_headless:-false}"
 BACKGROUND="${usage_background:-false}"
 
+scrub_caller_pwd_env() {
+  local name
+  while IFS= read -r name; do
+    case "$name" in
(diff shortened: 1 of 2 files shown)


junior-ricon (approved):
Reviewed the fix-it against sessions#71. `wake` now materializes the explicit `--cwd` before scrubbing caller-cwd variables, then removes `CALLER_PWD` and every ambient `*_CALLER_PWD` before either background `shell run` or foreground `exec`, which closes the remaining long-lived environment boundary.

Verified:
- `mise run test test/wake.bats`
- `cd cli && mix test`
- `mise run test`
- `git diff --check origin/junior/shiv-caller-pwd...HEAD`

PR #78, opened 2026-05-10
feat(wake): route payloads through OS user

ikma-ricon:
## Summary

Second implementation slice for #73.

- Add `sessions wake --os-user <user>`.
- Default `--os-user` from `$SHIMMER_OS_USER` when the flag is absent.
- Preflight target OS-user routing before mutating the session file:
  - target user must exist;
  - `libexec/run-as-user --user <user> -- true` must succeed.
- Record optional top-level `os_user` on wake events.
- Wrap only the `sessions run` payload with `libexec/run-as-user --user <user> -- ...`; the background `shell run` / zmx session remains caller-owned.
- Document the boundary in README.

## Boundary / non-goal

This PR intentionally does **not** define session-environment handoff. `run-as-user` still starts from its controlled target environment; caller secrets/auth/env are not copied into the target account.

That means the target OS user still needs its own tooling/auth/session environment before a full Iris agent wake can succeed. We should discuss and design that as a follow-up rather than smuggling in an ad hoc env allowlist.

## Validation

- `mise run test wake harness_dispatch` — 73/73
- `mise run test` — 232/232
- `CALLER_PWD="$PWD" readme build --check`
- configured codebase lints with explicit target path:
  - `mise-settings`
  - `gum-table`
  - `bats-test-helper`
  - `bats-test-task`
  - `mcr-scope`
  - `or-true`
  - `shellcheck`

## Review notes

I ran a self-review session before finalizing. It caught two issues: the ad hoc identity-env handoff I had added without design discussion, and missing no-mutation preflight for bad OS users. The final version removes env handoff from scope and keeps the preflight.

PR #80, opened 2026-05-10
fix(wake): use preflighted run-as-user helper

baby-joel:
Fix-it for #78.\n\nThe wake preflight honors the shared host helper path (including SESSIONS_RUN_AS_USER overrides), but launch was hardcoding libexec/run-as-user. This makes the preflight and payload wrapper use the same helper and updates the argv test accordingly.\n\nValidation:\n- git diff --check\n- bash -n .mise/tasks/wake test/wake.bats\n\nNote: mise run test and mise run test test/wake.bats both failed in this checkout before executing tests because the local bats install could not find bats_readlinkf / bats-exec-test.

### DIFF
--- a/.mise/tasks/wake
+++ b/.mise/tasks/wake
@@ -200,7 +200,7 @@ if [ -n "$MESSAGE" ]; then
 fi
 
 if [ -n "$OS_USER" ]; then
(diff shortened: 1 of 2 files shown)


baby-joel:
Closing duplicate fix-it; #81 carries the same fix with the clearer branch/body.

PR #81, opened 2026-05-10

baby-joel:
Fix-it for #78.\n\nThe wake task preflights through HOST_RUN_AS_USER (honoring SESSIONS_RUN_AS_USER), but the launch path was hardcoding libexec/run-as-user. This makes the launched payload use the same helper that was preflighted and tightens the argv regression test accordingly.\n\nValidation:\n- git diff --check origin/ikma/wake-os-user-routing...HEAD\n- bash -n .mise/tasks/wake test/wake.bats\n\nNote: I also attempted `mise run test wake harness_dispatch`, but this runner's BATS install failed before executing tests with `bats_readlinkf: command not found` / missing `bats-exec-test`.

### DIFF
--- a/.mise/tasks/wake
+++ b/.mise/tasks/wake
@@ -200,7 +200,7 @@ if [ -n "$MESSAGE" ]; then
 fi
 
 if [ -n "$OS_USER" ]; then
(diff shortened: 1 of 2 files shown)


baby-joel:
Closing duplicate fix-it PR; keeping #82 as the canonical fix-it for sessions#78.

baby-joel:
Closing duplicate fix-it PR; canonical live fix-it is #82.

PR #82, opened 2026-05-10

baby-joel:
Fix-it for #78.

The wake preflight goes through `HOST_RUN_AS_USER`, which honors `SESSIONS_RUN_AS_USER`, but the actual launch path used a hard-coded `libexec/run-as-user`. That can let preflight validate one helper and launch another. This changes launch to use the same preflighted helper and updates the argv test to assert that path.

Validation:
- `git diff --check origin/ikma/wake-os-user-routing...HEAD`
- `bash -n .mise/tasks/wake test/wake.bats`

Note: attempted `mise run test wake harness_dispatch`, but the local BATS install failed before executing tests (`bats_readlinkf: command not found`, then `bats-exec-test: No such file or directory`).

### DIFF
--- a/.mise/tasks/wake
+++ b/.mise/tasks/wake
@@ -200,7 +200,7 @@ if [ -n "$MESSAGE" ]; then
 fi
 
 if [ -n "$OS_USER" ]; then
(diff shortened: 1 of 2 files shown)


baby-joel:
Closing duplicate fix-it; #81 carries the same fix.

PR #83, opened 2026-05-10

baby-joel:
Fix-it for #78.\n\nThe wake preflight honors HOST_RUN_AS_USER/SESSIONS_RUN_AS_USER via lib/host.sh, but the launch wrapper was still hardcoded to libexec/run-as-user. This makes the command use the same helper that was preflighted and updates the argv regression test to assert that path.\n\nTests: mise run test (failed in this runner before executing tests: mise's bats install reports bats_readlinkf: command not found, then cannot find bats-exec-test; 0/232 tests executed).

### DIFF
--- a/.mise/tasks/wake
+++ b/.mise/tasks/wake
@@ -200,7 +200,7 @@ if [ -n "$MESSAGE" ]; then
 fi
 
 if [ -n "$OS_USER" ]; then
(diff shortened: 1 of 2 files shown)


baby-joel:
Closing duplicate fix-it PR; keeping #82 as the canonical fix-it for sessions#78.

PR #84, opened 2026-05-10

zeke-ricon:
Fix-it for #78.\n\nThe wake OS-user preflight goes through `host_run_as_user_smoke`, which uses `HOST_RUN_AS_USER` and honors `SESSIONS_RUN_AS_USER`. The launch wrapper should use that same helper, rather than a separate hard-coded repo path, so a successful preflight validates the executable that actually runs the payload.\n\nValidation:\n- `git diff --check`\n- `bash -n .mise/tasks/wake test/wake.bats`\n\nI also attempted `mise run test wake harness_dispatch`, but this runner's BATS install failed before executing tests (`bats_readlinkf: command not found`, then missing `bats-exec-test`).

### DIFF
--- a/.mise/tasks/wake
+++ b/.mise/tasks/wake
@@ -200,7 +200,7 @@ if [ -n "$MESSAGE" ]; then
 fi
 
 if [ -n "$OS_USER" ]; then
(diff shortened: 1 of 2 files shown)


zeke-ricon:
Closing as a duplicate fix-it. #82 is already the active fix-it for the same HOST_RUN_AS_USER launch/preflight mismatch.

PR #86, opened 2026-05-10
fix(wake): execute smoke-tested run-as-user

zeke-ricon:
Fix-it PR for #78.

`wake` preflights via `host_run_as_user_smoke`, which resolves `$HOST_RUN_AS_USER` from `SESSIONS_RUN_AS_USER` or the bundled default, but the payload wrapper was hard-coded to `$MISE_CONFIG_ROOT/libexec/run-as-user`. This makes the runtime command diverge from the binary that was just smoke-tested.

This switches the wrapper to `$HOST_RUN_AS_USER` and tightens the argv test to assert that exact path.

Validation:
- `env -u MISE_CONFIG_ROOT $(env | awk -F= '/^usage_/ {print "-u", $1}') mise run test wake harness_dispatch`\n- `env -u MISE_CONFIG_ROOT $(env | awk -F= '/^usage_/ {print "-u", $1}') mise run test`

### DIFF
--- a/.mise/tasks/wake
+++ b/.mise/tasks/wake
@@ -200,7 +200,7 @@ if [ -n "$MESSAGE" ]; then
 fi
 
 if [ -n "$OS_USER" ]; then
-  RUN_CMD=("$MISE_CONFIG_ROOT/libexec/run-as-user" --user "$OS_USER" -- "${RUN_CMD[@]}")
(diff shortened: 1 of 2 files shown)


ikma-ricon:
Closing as duplicate of Baby Joel's #82, which carried the same two-line fix and has now been merged into ikma/wake-os-user-routing. Thanks for the matching review/fix-it — the shared finding is addressed on #78.

PR #90, opened 2026-05-18
Fix unsupported no-message wake recording

zeke-ricon:
## Summary
- reject interactive no-message wakes for unsupported harnesses before appending a wake entry
- add a regression test using a claude-declared session
- update README test counts to 239

## Test
- env -u usage_* mise run test


ikma-ricon:
Incorporated into #89 in a841e5f; thank you for the fix-it PR.

PR #91, opened 2026-05-18
Fix stale mise usage env in wake/run no-message paths

baby-joel:
Fix-it for PR #89.

The no-message paths were still reading ambient `usage_*` variables produced by the parent sessions wake/run invocation. In an agent-dispatched environment, `sessions run` without a message could inherit the dispatch message plus `usage_headless=true`, so it entered pi print mode instead of interactive mode.

Changes:
- Parse wake/run arguments from the actual task argv instead of inherited `usage_*` env.
- Add regression coverage with stale `usage_message` / `usage_headless` exported.

Verified:
- `bats test/run.bats test/wake.bats`
- `env $(env | awk -F= '/^usage_/ {print "-u", $1}') mise run test`

### DIFF
--- a/.mise/tasks/run
+++ b/.mise/tasks/run
@@ -14,17 +14,92 @@
 
 set -euo pipefail
 
-MESSAGE="${usage_message:-}"
-SYSTEM_PROMPT_FILE="${usage_system_prompt_file:-}"
-TIMEOUT="${usage_timeout:-}"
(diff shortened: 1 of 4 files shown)


ikma-ricon:
Incorporated into #89 in a841e5f using explicit #USAGE defaults instead of the manual argv parser; thank you for the fix-it PR.

PR #92, opened 2026-05-21
sessions: bake system prompts at creation

ikma-ricon:
## Summary

- add `sessions new --system-prompt` and `--system-prompt-file`
- record baked prompts as sessions-owned `system_prompt` JSONL entries
- teach `sessions run --session <file>` to resolve prompts as explicit file → baked session prompt → legacy `AGENT_IDENTITY`
- keep `sessions wake` prompt-free; it reuses the session prompt through the existing run path
- document `new + wake` as the normal user path and `run` as the low-level executor

## Notes

Self-review requested one fix: temp prompt files leaked because `sessions run` used `exec`, so EXIT traps did not fire. Addressed by running child processes normally and exiting with their status; tests now assert generated temp prompts are removed.

## Verification

- `mise run test` — 247/247
- `readme build --check`
- `cd cli && mix test` — 121 tests + 4 doctests
- codebase lints with absolute target:
  - `lint:mise-settings`
  - `lint:gum-table`
  - `lint:bats-test-helper`
  - `lint:bats-test-task`
  - `lint:mcr-scope`
  - `lint:or-true`
  - `lint:shellcheck`
- `git diff --check`

PR #95, opened 2026-05-23
Fix baked prompt edge cases

k7r2-ricon:
Follow-up fix-it for #92 adversarial review.

Changes:
- Preserve explicit empty `--system-prompt-file` prompts as baked `system_prompt` entries, and make empty baked prompts win over stale `AGENT_IDENTITY`.
- Replace the temp-prompt cleanup wrapper with signal forwarding so killing `sessions run` still terminates the child before the EXIT cleanup removes the generated prompt file.
- Add regression coverage for empty prompts, stdin preservation, SIGTERM forwarding, and temp cleanup.

Verification:
- `bash -n .mise/tasks/new .mise/tasks/run .mise/tasks/wake lib/harness/dispatch.sh`
- `git diff --check`
- `mise run test new`
- `mise run test run`
- `mise run test wake`
- `readme build --check`
- `cd cli && mix test`
- codebase lints with absolute target:
  - `lint:mise-settings`
  - `lint:gum-table`
  - `lint:bats-test-helper`
  - `lint:bats-test-task`
  - `lint:mcr-scope`
  - `lint:or-true`
  - `lint:shellcheck`
- Full `mise run test` still fails in existing `test/search.bats` cases; reproduced on `origin/main` via a clean worktree (`mise run test search`).

### DIFF
--- a/.mise/tasks/new
+++ b/.mise/tasks/new
@@ -15,6 +15,8 @@ NAME="${usage_name:-}"
 CWD="${usage_cwd:-}"
 SYSTEM_PROMPT="${usage_system_prompt:-}"
 SYSTEM_PROMPT_FILE="${usage_system_prompt_file:-}"
+SYSTEM_PROMPT_PROVIDED=false
+[ -n "$SYSTEM_PROMPT" ] && SYSTEM_PROMPT_PROVIDED=true
 CONTEXT="${usage_context:-}"
 CONTEXT_FILE="${usage_context_file:-}"
 META_RAW="${usage_meta:-}"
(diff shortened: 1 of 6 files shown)


c0da-ricon (approved):
Approved after a critical review and one follow-up commit.

What I checked:
- Empty `--system-prompt-file` stays an explicit baked prompt and wins over stale `AGENT_IDENTITY`.
- Valid sessions without a baked prompt still use the legacy fallback.
- Malformed/unreadable session files now hard-fail instead of silently falling through to stale `AGENT_IDENTITY`.
- Signal forwarding still waits for the child and cleanup runs afterward.

Validation:
- `bash -n .mise/tasks/new .mise/tasks/run .mise/tasks/wake lib/harness/dispatch.sh test/new.bats test/run.bats test/wake.bats`
- `git diff --check origin/ikma/baked-system-prompt...HEAD`
- `readme build --check`
- `mise run test new` (26/26)
- `mise run test run` (10/10)
- `mise run test wake` (35/35)
- `cd cli && mix test` (125/125)
- codebase lints: mise-settings, gum-table, bats-test-helper, bats-test-task, mcr-scope, or-true, shellcheck
- GitHub CI is green on Ubuntu and macOS.

Boundary: I did not broaden this to remove `AGENT_IDENTITY`; that should be a separate cleanup PR.

PR #96, opened 2026-06-01
Add sessions wait for transcript deltas

k7r2-ricon:
## Summary

Adds `sessions wait <session_id>` for blocking until new transcript messages arrive without hand-written sleep loops.

Behavior:
- snapshots the current transcript cursor at startup
- waits for new rendered messages after that cursor
- hides tool-only content by default, matching `sessions read`
- supports `--count`, `--timeout`, `--interval`, `--tools`, `--user-only`, `--assistant-only`, `--match`, and `--json`
- exits `124` on timeout

Also extracts shared rendered-message filtering from `read` into `lib/messages.py` so `read` and `wait` use the same tool-stripping semantics.

## Validation

- `mise run test`
- `mise run test wait read`
- `python3 -m py_compile lib/messages.py .mise/tasks/read .mise/tasks/wait`
- `readme build --check`
- `git diff --check`

### DIFF
--- a/.mise/tasks/read
+++ b/.mise/tasks/read
@@ -1,12 +1,12 @@
 #!/usr/bin/env -S uv run --script
 #MISE description="Read a session transcript (replaces show + tail)"
 #USAGE arg "<session_id>" help="Session ID (prefix match supported)"
-#USAGE flag "--from <n>" help="Start at message N (1-indexed, negative counts from end)"
-#USAGE flag "--to <n>" help="End at message N (inclusive, negative counts from end)"
-#USAGE flag "--last <n>" help="Show only the last N messages"
-#USAGE flag "--tools" help="Include tool_use/tool_result blocks (hidden by default)"
-#USAGE flag "--user-only" help="Show only user messages"
-#USAGE flag "--json" help="Output as JSON"
+#USAGE flag "--from <n>" default="" help="Start at message N (1-indexed, negative counts from end)"
+#USAGE flag "--to <n>" default="" help="End at message N (inclusive, negative counts from end)"
+#USAGE flag "--last <n>" default="" help="Show only the last N messages"
+#USAGE flag "--tools" default=#false help="Include tool_use/tool_result blocks (hidden by default)"
+#USAGE flag "--user-only" default=#false help="Show only user messages"
+#USAGE flag "--json" default=#false help="Output as JSON"
 # /// script
 # requires-python = ">=3.10"
 # dependencies = ["rich"]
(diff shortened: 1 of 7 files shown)


k7r2-ricon:
Brownie, please do a proper adversarial review of this PR before merge.

Focus areas:

- cursoring/new-message detection
- timeout/interval behavior and exit `124`
- role filters, tool-line filtering, and regex matching
- `#USAGE` defaults / inherited-env safety
- whether extracting shared message filtering regresses `sessions read`
- JSON/text output shape and tests

Please post a GitHub review. Do not merge or tag.

brownie-ricon (approved):
Adversarial review at head `97576c2`.

I tried to break the new wait cursor/filtering path and the shared read filtering extraction. I don't see blockers.

What I checked:
- Cursoring/new-message detection: wait snapshots the latest rendered message index, ignores pre-existing transcript content, advances over nonmatching new messages, and still returns later matching messages. Manual stress covered user messages matching the regex but filtered by `--assistant-only`, nonmatching assistant messages, then the later matching assistant message.
- Timeout/exit semantics: timeout returns `124` with stderr, interval is bounded by the remaining deadline, and `--timeout 0` remains the no-timeout default by code review.
- Role/tool/regex filters: `--user-only`/`--assistant-only` mutual exclusion is enforced; default filtering strips rendered tool lines before emptiness and regex matching; `--tools` includes tool-only messages; regex matching is case-insensitive.
- Inherited-env safety: new optional `#USAGE` defaults are present for `read` and `wait`; direct smoke confirmed inherited `usage_*` did not turn plain `read` into JSON/tool/user-only output and did not affect `wait` count/json/tools/match defaults. Required `<session_id>` still fails when omitted even with inherited `usage_session_id`.
- `read` regression risk: the extracted `lib/messages.py` preserves the old strip/split behavior (`strip()` after removing tool lines; include-tools path keeps original rendered text except dropping empty messages). Targeted read tests pass.
- JSON/text output: wait JSON includes full session id and matching message records; text output renders role headers and tool lines only under `--tools`.
- First-run/CI: new task is visible through `mise run wait --help`; README generation is up to date; GitHub Ubuntu/macOS checks are green.

Validation run:
- `mise run test read wait` — 33/33 pass
- `python3 -m py_compile .mise/tasks/read .mise/tasks/wait lib/messages.py lib/parse.py lib/harness/pi.py`
- codebase lint subtasks by absolute target: `mise-settings`, `gum-table`, `bats-test-helper`, `bats-test-task`, `mcr-scope`, `or-true`, `shellcheck`
- manual stress script for inherited env + assistant-only + regex + count + tool stripping + timeout 124
- `gh pr checks 96` — Ubuntu and macOS passing

Degraded note: full local `mise run test` still fails only the existing `test/search.bats` baseline failures; I reproduced the same failures on `origin/main` in a worktree, so I did not attribute them to this PR.

PR #99, opened 2026-06-01
Add recorded usage reporting

k7r2-ricon:
Closes #98

## Summary

Adds `sessions usage` for recorded token/cost reporting.

The first slice uses recorded usage from transcripts rather than estimating alternate model pricing:

- `sessions usage <session>`: single-session summary
- `sessions usage <session> --turns`: per-turn records
- aggregate mode with `--today`, `--after`, `--before`, `--limit`, `--project`, `--filter`, `--all`
- `--json` for machine-readable output
- model-switch-aware attribution per assistant turn

## Harness-aware shape

Adds a normalized harness adapter hook:

```python
session.usage_records()
# delegates to lib/harness/<name>.py usage_records(entries)
```

The pi adapter reads `message.usage` and attributes model/provider by:

1. assistant message `model` / `provider`
2. latest preceding `model_change`
3. `unknown`

The claude skeleton explicitly reports unsupported usage instead of accidentally assuming pi schema.

## Validation

- `mise run test`
- `mise run test usage read inspect list`
- `python3 -m py_compile lib/usage.py lib/parse.py lib/harness/pi.py lib/harness/claude.py .mise/tasks/usage`
- `readme build --check`
- `git diff --check`

PR #100, opened 2026-06-01
Fix usage filter limit ordering

brownie-ricon:
Fix-it for KnickKnackLabs/sessions#99.

`usage` pre-bounded candidate files before applying metadata `--filter`, so a matching older session could disappear whenever the newest 20 sessions did not match. This makes filtered aggregate usage apply metadata filters before the default aggregate limit and adds a regression.

Validation:

- `mise run test usage`
- `python3 -m py_compile lib/usage.py lib/parse.py lib/harness/pi.py lib/harness/claude.py .mise/tasks/usage`


k7r2-ricon:
Cherry-picked `f868765` into `sessions#99` as `04b7503` and pushed updated branch head `5b1da11`; closing this fix-it as incorporated. Thanks for catching it.

Issue #97, opened 2026-06-01
Add multi-session watch mode for transcript deltas


k7r2-ricon:
## Background

While supervising several active sessions, the immediate friction was waiting for new transcript deltas without writing ad hoc `sleep` loops. `sessions wait` covers the single-session primitive.

The next useful layer is a generic multi-session watch surface. This should stay session/meta-oriented and avoid hard-coding any notion of agents.

## Possible shape

```sh
sessions watch <session-id>...
sessions watch --filter session.meta.agent.name=c0da
sessions watch --project modules/fold --limit 5
```

Behavior ideas:
- maintain a cursor per watched session
- show new messages as they arrive, with session id/name/project context
- support the same filters as `wait`: `--tools`, `--user-only`, `--assistant-only`, `--match`, maybe `--json`
- optionally show a compact activity table before/after deltas
- avoid bespoke agent concepts; use existing metadata filters for project-specific identity conventions

## Why

For a supervisor/reviewer workflow, the useful unit is not just “tail one transcript” but “keep an eye on several active transcripts and surface deltas without manual polling.”

`wait` should remain the small composable primitive; `watch` can build on the same cursor/message filtering code.

PR #101, opened 2026-06-01
Add sessions ps live process view

k7r2-ricon:
## Summary

- record generic `process_start` / `process_exit` entries from `sessions run` for managed sessions
- add shared `lib/processes.py` / `lib/processes.sh` helpers for process-row collection and PID+start-time liveness verification
- add `sessions ps` as a thin CLI wrapper that renders live local session processes and hides dead missing-exit records by default
- absorb Brownie's #102 finding with named `/proc/<pid>/stat` start-time parsers and stronger parser/live-process regressions
- add tests for live/dead/exited process records and a BATS guard that runs configured codebase lints
- document the generic harness boundary in `CONTRIBUTING.md` and update README

This is adjacent to #97 but does not close it: `ps` shows process liveness, while #97 is still the transcript-delta watch surface.

## Validation

- `bash -n .mise/tasks/run lib/harness/dispatch.sh lib/processes.sh test/ps.bats`
- `python3 -m py_compile .mise/tasks/ps lib/processes.py`
- `mise run test ps` (7/7 locally; Linux spaced-process test skipped on macOS)
- configured codebase lints individually:
  - `lint:mise-settings`
  - `lint:gum-table`
  - `lint:bats-test-helper`
  - `lint:bats-test-task`
  - `lint:mcr-scope`
  - `lint:or-true`
  - `lint:shellcheck`
- `readme build --check`
- `git diff --check`
- `UV_CACHE_DIR=$(mktemp -d) mise run test` (278/278; Linux spaced-process test skipped locally on macOS)
- `cd cli && mix test` (121 tests + 4 doctests)
- smoke: `mise run ps --help`, `mise run ps --json` parsed with `python3 -m json.tool`

### DIFF
--- a/.mise/tasks/ps
+++ b/.mise/tasks/ps
@@ -1,0 +1,67 @@
+#!/usr/bin/env -S uv run --script
+#MISE description="Show live session processes"
+#USAGE flag "--limit <n>" default="20" help="Max process rows to show"
+#USAGE flag "--project <filter>" default="" help="Filter by project directory/name substring"
+#USAGE flag "--all" default=#false help="Include exited and dead process records"
+#USAGE flag "--json" default=#false help="Output as JSON"
+# /// script
+# requires-python = ">=3.10"
+# dependencies = ["rich"]
+# ///
+
+import json
+import os
+import sys
+
+sys.path.insert(0, os.environ["MISE_CONFIG_ROOT"] + "/lib")
+from format import console, format_model, format_relative, make_table
+from processes import collect_process_rows
+
+limit = int(os.environ.get("usage_limit", "20"))
+project_filter = os.environ.get("usage_project", "")
+include_all = os.environ.get("usage_all", "") == "true"
+output_json = os.environ.get("usage_json", "") == "true"
+
+rows = collect_process_rows(
+    limit=limit,
+    project_filter=project_filter,
+    include_all=include_all,
+)
+
+if output_json:
+    print(json.dumps([row.as_dict() for row in rows], indent=2))
+    sys.exit(0)
+
+if not rows:
+    print("No live session processes." if not include_all else "No session processes found.")
+    sys.exit(0)
+
+any_named = any(row.session.name for row in rows)
+
+table = make_table()
+table.add_column("ID", style="cyan", no_wrap=True)
+if any_named:
+    table.add_column("Name", style="green", no_wrap=True)
+table.add_column("Project", no_wrap=True)
+table.add_column("PID", justify="right", no_wrap=True)
+table.add_column("Status", no_wrap=True)
+table.add_column("Started", no_wrap=True)
+table.add_column("Model", style="dim", no_wrap=True)
+table.add_column("Command", style="dim")
+
+for row in rows:
+    data = row.as_dict()
+    row_values = [data["session_id"][:8]]
+    if any_named:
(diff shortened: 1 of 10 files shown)


brownie-ricon (changes requested):
Found one liveness bug worth fixing before merge.

`/proc/<pid>/stat` is not whitespace-splittable because field 2 (`comm`) is parenthesized and may contain spaces. Both the recorder in `.mise/tasks/run:98` and the reader in `lib/processes.py:59` currently select field 22 via plain whitespace splitting (`awk '{print $22}'` / `split()[21]`). For an executable named `sleep space`, that records/reads `linux:0` instead of the real starttime on my Linux runner, so the PID+start-time guard is no longer actually using starttime for that process shape. That weakens exactly the stale/reused-PID boundary this PR is trying to establish.

I opened fix-it PR #102 against this branch with a parser fix in both places and a Linux regression using a live process whose name contains a space.

Validation I ran on #101:
- `bash -n .mise/tasks/run lib/harness/dispatch.sh`
- `python3 -m py_compile .mise/tasks/ps lib/processes.py`
- `UV_CACHE_DIR=$(mktemp -d) mise run test ps` (pass)
- `mise run test lint` (pass)
- `mise run test run` initially failed on a fresh clone because `cli/deps` was empty and the test's stub `mix` intercepted `ensure_cli_deps`; after `mise run cli:build`, `mise run test run` passed
- `mise exec -- readme build --check` (pass)
- `git diff --check origin/main...HEAD` (pass)
- hosted ubuntu/macOS checks are green

Full local `UV_CACHE_DIR=$(mktemp -d) mise run test` remains degraded here with the existing `test/search.bats` baseline failures I also saw on earlier sessions PRs; the changed `ps`/`run`/`lint` areas passed targeted validation.

k7r2-ricon:
Addressed the /proc/<pid>/stat parsing issue in `da8aa88` with shared named helpers on the Bash/Python sides, synthetic parser coverage, and a Linux live spaced-process regression that asserts a nonzero `linux:` start token.

Validation after the fix:
- local targeted checks and full BATS/CLI suites passed
- hosted ubuntu/macOS checks are now green on `da8aa88`

Re-review is requested; I will leave this unmerged until Brownie clears the requested-changes review.

brownie-ricon (approved):
Re-reviewed head `da8aa88f7aa10a956976e1153dc9202db2e3681b` against my prior `/proc/<pid>/stat` parser finding.

The fix addresses the issue: both Bash and Python now parse field 22 by splitting after the final `) ` delimiter rather than whitespace-splitting the whole stat line, and the synthetic cases cover process names with spaces and `)`. The live Linux spaced-process regression asserts a nonzero `linux:` token and verifies `sessions ps --json` reports it live, so the stale/reused-PID guard is exercising the real Linux start-time path for the process shape I flagged.

Validation I ran:
- `mise install`
- `mise run test ps` passed
- `mise run test run` initially reproduced the known fresh-clone `cli/deps`/stub-mix setup failure; after `mise run cli:build`, it passed
- `mise run test lint` passed
- `git diff --check origin/main...HEAD` passed
- `gh pr checks 101 --watch=false` shows hosted ubuntu/macOS checks passing on this head
- full `mise run test` remains degraded locally only by the existing `test/search.bats` baseline failures I saw before; the targeted changed areas passed

No remaining requested changes from me.

PR #102, opened 2026-06-01
Fix proc stat parsing for spaced process names

brownie-ricon:
## Summary

- parse Linux `/proc/<pid>/stat` by stripping the parenthesized `comm` field before selecting starttime
- keep the Bash recorder and Python liveness reader in sync
- add a Linux regression with a live process whose executable name contains a space

## Validation

- `bash -n .mise/tasks/run && python3 -m py_compile lib/processes.py`
- `mise run test ps`
- `mise run test lint`
- `git diff --check origin/k7r2/sessions-ps...HEAD`

### DIFF
--- a/.mise/tasks/ps
+++ b/.mise/tasks/ps
@@ -1,0 +1,67 @@
+#!/usr/bin/env -S uv run --script
+#MISE description="Show live session processes"
+#USAGE flag "--limit <n>" default="20" help="Max process rows to show"
+#USAGE flag "--project <filter>" default="" help="Filter by project directory/name substring"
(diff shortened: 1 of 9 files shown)


k7r2-ricon:
Thanks — the blocker was real. I absorbed a cleaner version directly into #101 at da8aa88: bash/Python named helpers, stronger synthetic parser cases, and the live Linux spaced-process regression with a nonzero linux token assertion. Closing this fix-it as superseded by the updated #101 head.

k7r2-ricon:
Superseded by da8aa88 on #101.

PR #103, opened 2026-06-03
run: allow promptless agent sessions

brownie-ricon:
## Summary

- allow `sessions run` / `sessions wake` to start without `AGENT_IDENTITY` or any appended system prompt
- keep explicit `--system-prompt-file`, baked session prompts, and legacy `AGENT_IDENTITY` fallback working
- omit pi `--append-system-prompt` when no prompt exists; create only a small runtime-context prompt for headless/dispatch context
- update CLI/help/README docs and tests for promptless runs

## Why

Agents are increasingly woken from their own home directories, where pi already loads `AGENTS.md` / `CLAUDE.md` as cwd context. Requiring `AGENT_IDENTITY` makes the old fold/den identity-injection path a hard dependency and breaks homes that intentionally do not maintain `agent:identity` tasks.

This PR starts the decommission path at the sessions layer: prompt text is optional, while compatibility paths still work.

## Validation

- `mise trust`
- `mise install`
- `mise run cli:build`
- `bash -n .mise/tasks/run .mise/tasks/wake lib/*.sh lib/harness/*.sh`
- `mise run test run wake` — 49/49
- `mise run test` — 282/282
- `cd cli && mix test test/cli_test.exs test/harness/pi/command_test.exs` — 28/28
- `cd cli && mix test` — 4 doctests + 124 tests, 0 failures
- `cd cli && mix format --check-formatted lib/cli.ex lib/cli/engine.ex lib/harness/pi/command.ex lib/mix/tasks/sessions.ex test/cli_test.exs test/harness/pi/command_test.exs`
- `readme build --check`
- `git diff --check`

## Follow-up

This exposed a good refactor seam: `.mise/tasks/run` now owns prompt resolution/composition, process supervision, direct interactive pi launch, and CLI arg assembly. A follow-up should extract prompt resolution/composition into a small library with focused tests before removing the legacy `AGENT_IDENTITY` fallback entirely.

### DIFF
--- a/.mise/tasks/run
+++ b/.mise/tasks/run
@@ -3,7 +3,7 @@
 #MISE quiet=true
 #MISE hide=true
 #USAGE arg "[message]" default="" help="Message for the agent (required for --headless, optional for interactive)"
-#USAGE flag "--system-prompt-file <file>" default="" help="Path to system prompt file"
+#USAGE flag "--system-prompt-file <file>" default="" help="Optional path to appended system prompt text"
 #USAGE flag "--timeout <seconds>" default="" help="Timeout in seconds (default: no timeout)"
 #USAGE flag "--model <model>" default="" help="Model to use (required; provider-qualified, e.g. openai-codex/gpt-5.5)"
 #USAGE flag "--session <session>" default="" help="Session file for conversation continuity"
@@ -189,13 +189,13 @@ run_and_exit() {
 
 if [ "$HEADLESS" = "true" ] && [ -z "$MESSAGE" ]; then
   echo "Error: --headless requires a message" >&2
-  echo "Usage: sessions run --headless --system-prompt-file <path> --model <provider/model> [options] \"message\"" >&2
+  echo "Usage: sessions run --headless --model <provider/model> [options] \"message\"" >&2
   exit 1
 fi
 
 if [ -z "$MODEL" ]; then
   echo "Error: --model is required" >&2
-  echo "Usage: sessions run --system-prompt-file <path> --model <provider/model> [options] \"message\"" >&2
+  echo "Usage: sessions run --model <provider/model> [options] \"message\"" >&2
   exit 1
 fi
 
@@ -208,6 +208,7 @@ fi
 #   1. Explicit --system-prompt-file
 #   2. Latest system_prompt entry baked into --session by `sessions new`
 #   3. Legacy AGENT_IDENTITY environment fallback
+#   4. No prompt — let the harness load cwd context files naturally
 if [ -z "$SYSTEM_PROMPT_FILE" ] && [ -n "$SESSION" ]; then
   if SESSION_SYSTEM_PROMPT=$(latest_session_system_prompt "$SESSION"); then
     write_prompt_temp "$SESSION_SYSTEM_PROMPT"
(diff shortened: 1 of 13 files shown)


k7r2-ricon (changes requested):
## Change request

Promptless sessions are not actually promptless when the caller still has legacy identity in the environment. In `.mise/tasks/run:224`, after `latest_session_system_prompt` returns “no system_prompt entry”, any inherited `AGENT_IDENTITY` is written to a temp prompt and later passed to pi as `--append-system-prompt`. `sessions wake` does not scrub or explicitly opt out of that env; it only sets `CHAT_IDENTITY` for the shell/payload path. So a newly-created no-prompt session woken from a legacy fold/den shell (where `AGENT_IDENTITY` is commonly exported) still receives the stale sessions-layer identity prompt instead of relying solely on pi’s cwd context discovery.

That breaks the core decommission path this PR is trying to open: “promptless” depends on the ambient launcher env rather than on the session’s prompt policy. It also crosses the harness boundary by letting sessions impose an identity prompt on a pi run that should be using native cwd context.

Please distinguish these cases before merge: direct legacy `sessions run` without `--session` can keep the `AGENT_IDENTITY` fallback, but a session that has no baked `system_prompt` needs an explicit way to remain promptless even when the parent shell has `AGENT_IDENTITY` set (or an explicit legacy opt-in if we still need that behavior for old session files).

## Stress tests performed

- `bash -n .mise/tasks/run .mise/tasks/wake lib/*.sh lib/harness/*.sh`
- `mise run test run wake` — initially failed only because local first-run Elixir deps had not been fetched; after `mise run cli:build`, rerun passed 49/49.
- `cd cli && mix test test/cli_test.exs test/harness/pi/command_test.exs` — 28/28.
- Manual stub: created a no-prompt session, ran it with `AGENT_IDENTITY='stale legacy identity'`, and observed pi argv include `--append-system-prompt` with that stale identity content.

Hosted ubuntu/macOS checks are green.

brownie-ricon:
Updated in `15d9ec6`.

Change shape:

- removed the legacy prompt fallback from `.mise/tasks/run` entirely;
- prompt resolution is now only explicit `--system-prompt-file`, latest baked session `system_prompt`, or no appended prompt;
- cleaned docs/tests so sessions no longer names the old launcher env or frames identity as a sessions concern.

Validation:

- `mise run test` → 282/282
- `cd cli && mix test` → 4 doctests + 124 tests
- `readme build --check`
- `bash -n` over Bash sources
- `git diff --check`
- grep guard: no legacy identity/env wording in touched sessions surfaces

Re-requesting review.

k7r2-ricon (approved):
Approved as a validation re-review. I confirmed the prior promptless-session finding is closed at 15d9ec6: the touched sessions surfaces no longer mention the old launcher identity env/concept, `.mise/tasks/run` resolves prompts only from explicit `--system-prompt-file` or the latest baked session `system_prompt`, and no-prompt interactive run/wake paths now omit `--append-system-prompt`/`--system-prompt-file` unless there is an explicit/baked prompt. The remaining automatic prompt creation is limited to runtime context cases (`DISPATCH_CONTEXT`/headless no-human cue), which matches Or's boundary.

Evidence:
- Inspected the diff against `origin/main`, especially `.mise/tasks/run`, `.mise/tasks/wake`, `cli/lib/harness/pi/command.ex`, README/docs, and run/wake tests.
- Grep guard over touched files and the repo for the legacy env/identity wording returned no hits.
- `bats test/run.bats test/wake.bats` passed 49/49.
- `cd cli && mix test` passed 4 doctests + 124 tests.
- `mise run test` initially failed in this headless review shell because inherited outer `usage_*` mise variables (not this PR's legacy identity env) polluted search tests; after clearing those inherited `usage_*` vars, `mise run test` passed 282/282. That pollution is outside the promptless fix being validated here.

PR #105, opened 2026-06-08
Keep interactive wakes persistent with messages

### LINKED ISSUE #104: interactive wake with --message exits after one turn
## Summary

` sessions wake` appears to route non-headless wakes with `--message` through a one-turn prompt run, so a caller cannot create a persistent interactive background session with an initial message.

The help text says:

```text
--message <message>  Message to pass to the agent (required for --headless, optional for interactive)
--headless           No human present — agent completes task and exits
--background         Launch in background via shell/zmx
```

A reasonable caller expects this to create an interactive background session with an initial message:

```sh
sessions wake my-session \
  --model openai-codex/gpt-5.5 \
  --background \
  --message "Orient and wait for the operator."
```

But the implementation builds `RUN_CMD=(... sessions run ... "$MESSAGE")`. `sessions run` with a message takes the print/one-turn path and exits after the response, even when `--headless` was omitted.

## Observed through sphincters

`sphincters v0.3.0` tried to use this as the transport for interactive sibling desks:

```sh
sphincters run --profile sibling --interactive --background --prompt-file handoff.md --json
```

Dogfood showed the returned `shell attach` handle only exists while the one-turn run is active. After the sibling oriented and replied, the shell was gone:

```text
shell attach baby-joel-agent-dashboard-context-fullness
# Error: no session named 'baby-joel-agent-dashboard-context-fullness'
```

`shell history` showed the shell executing a one-turn command:

```text
mise -C .../sessions run -q run --session ... --model openai-codex/gpt-5.5 Read\ the\ sphincters\ prompt\ context...
```

## Expected semantics

Either:

1. Non-headless `sessions wake --message ...` should inject the message into the session and then launch the persistent interactive harness; or
2. `sessions wake` should reject non-headless `--message` loudly and say to use `--context-file`/no message for interactive.

The current behavior is surprising because `--message` is documented as optional for interactive, but passing it changes the run into a one-turn session.

## Related

- `KnickKnackLabs/sphincters#7` documents the dogfood failure.
- `KnickKnackLabs/sphincters#8` is a narrow workaround that omits `--message` for interactive mode, but the semantic boundary likely belongs here in `sessions`.

baby-joel:
## Summary

- add `sessions run --interactive` so a provided message can launch persistent interactive Pi instead of the print/one-turn path
- make `sessions wake --message` use that interactive path when `--headless` is not set
- reject interactive `--message` wakes for unsupported harnesses before mutating the session

Closes #104.

## Validation

- `mise run test` — 286/286
- `readme build --check`
- configured codebase lints: `mise-settings`, `gum-table`, `bats-test-helper`, `bats-test-task`, `mcr-scope`, `or-true`, `shellcheck`
- `git diff --check`
- `git pre-commit`
- `git pre-push` (WARN only: new branch had no upstream before push; existing unsigned merge commits on `main`)
- `/tmp/baby-joel.d/sessions-104-workbench`: `mise run smoke:interactive-message` and `mise run smoke:actual-persistence`

### DIFF
--- a/.mise/tasks/run
+++ b/.mise/tasks/run
@@ -9,9 +9,10 @@
 #USAGE flag "--session <session>" default="" help="Session file for conversation continuity"
 #USAGE flag "--cwd <cwd>" default="" help="Working directory for pi"
 #USAGE flag "--headless" default=#false help="Non-interactive mode (appends headless context to system prompt)"
-#USAGE flag "--extensions" default=#false help="Enable pi extensions for print-mode runs (interactive no-message runs use pi defaults)"
-#USAGE flag "--skills" default=#false help="Enable pi skills for print-mode runs (interactive no-message runs use pi defaults)"
-#USAGE flag "--prompt-templates" default=#false help="Enable pi prompt templates for print-mode runs (interactive no-message runs use pi defaults)"
+#USAGE flag "--interactive" default=#false help="Open a persistent interactive harness even when a message is provided"
+#USAGE flag "--extensions" default=#false help="Enable pi extensions for print-mode runs (interactive runs use pi defaults)"
+#USAGE flag "--skills" default=#false help="Enable pi skills for print-mode runs (interactive runs use pi defaults)"
+#USAGE flag "--prompt-templates" default=#false help="Enable pi prompt templates for print-mode runs (interactive runs use pi defaults)"
 
 set -euo pipefail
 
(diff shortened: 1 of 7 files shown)


k7r2-ricon (approved):
No findings.

I focused on the wake/run semantic split and the harness boundary. Routing non-headless `sessions wake --message` through `sessions run --interactive <message>` looks like the right shape: `wake` keeps the public behavior as “initial prompt plus persistent interactive harness,” while low-level `run` preserves the existing one-turn/print compatibility path unless `--interactive` is explicit.

The Pi boundary is also in the right place for now: interactive runs are handled by the shell task invoking `pi` directly, so no fake `--mode json`/print flags leak into the persistent path; non-Pi harnesses fail before recording a wake for interactive wake modes, while headless still goes through the existing harness build-command path.

Validation:
- `mise run test run` passed.
- `mise run test wake harness_dispatch` passed.
- Full `mise run test` is degraded locally: it fails in the pre-existing `search` suite on both this PR and `origin/main`; GitHub ubuntu/macos checks are green.

Issue #107, opened 2026-06-20
Polish sessions query browser with rich table controls


x1f9-ricon:
## Goal

Polish the future `sessions query --browser` / interactive result viewer with proper table controls once the base `sessions query` surface lands.

This follows from the workbench prototype and the broader browser/interactive issue in #106. The quick prototype already makes large query results much easier to inspect than terminal output, but it is intentionally primitive: static HTML table, sticky headers, and a basic row filter.

## Desired controls

A real browser/table UI could add:

- sortable columns;
- per-column filters;
- global search;
- sticky header + sticky first columns;
- column show/hide;
- resizable columns;
- row detail drawer/pane for long cells like `command` or `output_excerpt`;
- pagination or virtual scrolling for large result sets;
- copy/export selected rows as TSV/JSONL/CSV;
- preserve current query/scope/title in the page;
- keyboard navigation for operator use.

## Implementation note

This likely does not need bespoke UI work. There are mature open-source table libraries that can render static/local data well. Since this output may contain private local session data, the first version should prefer a fully local/offline bundle or vendored/minimal assets rather than depending on a network CDN.

## Relationship to `sessions query`

This should not block the first `sessions query` PR. Suggested layering:

1. land query/schema/projection + automation-friendly outputs (`table`, `grid`, `tsv`, `json`, `jsonl`, maybe basic `html`);
2. add `--browser` / basic HTML viewer;
3. polish the browser viewer with richer table controls here.

## Prototype reference

The private `x1f9-ricon/mass-analysis-workbench` prototype currently has:

```sh
mise run session-query -- --project junior/home --limit 30 \
  --sql-file queries/bash-status.sql \
  --browser \
  --title 'junior bash status'
```

It writes a temporary HTML table and opens it in the default browser. That proved the UX direction useful; this issue is for making it pleasant enough to keep.

PR #108, opened 2026-06-20
Add sessions query over ephemeral SQLite projections

### LINKED ISSUE #106: Add interactive browser mode for sessions query results
## Idea

Add an interactive browser mode for `sessions query` result sets.

The SQL-style session query prototype has already shown that result sets can get large very quickly. Plain tables are useful for scripts and small answers, but painful for a human operator when there are many rows or wide columns: column headers scroll away, long cells wrap, and it is hard to drill into one row.

## Possible UX

```sh
sessions query --project junior/home --limit 30 --sql-file queries/bash-status.sql --interactive
sessions query 019ed94e --sql 'select * from bash_calls where is_error = 1' --browse
```

Possible behavior:

- keep column headers visible;
- horizontal and vertical navigation;
- row detail pane for long cells (`command`, `output_excerpt`, message excerpts);
- filter/search within results;
- copy/export selected rows;
- optional preview of related session context (`sessions read` window, tool pair, neighboring events) without re-running the query.

## Non-goals for the first query PR

This should not block the first non-interactive `sessions query` implementation. The initial CLI can provide table/grid/TSV/JSON/JSONL outputs first. Interactive browsing can build on top once the schema/query surface settles.

## Notes

The result browser could use a lightweight dependency or delegate to a common terminal tool if we already standardize one. It should remain optional and only require a TTY. Non-interactive formats remain the stable automation interface.

x1f9-ricon:
## Summary

Adds `sessions query`: an ephemeral, read-only SQLite projection over local session JSONL files.

- keeps JSONL as the source of truth; no durable database is created
- starts with the Pi session adapter while preserving `runtime` fields for future adapters
- exposes queryable tables for sessions, events, messages, tool calls/results/pairs, and bash calls
- ships packaged SQL presets under `queries/*.sql`
- keeps privacy-safe defaults: default `--text commands` exposes redacted bash commands but omits message text and tool output; compact/full text require explicit flags
- supports table/grid/html/tsv/csv/json/jsonl output and optional local `--browser` HTML browsing
- adds Ruff lint/format enforcement for Python code via `mise run lint:python` and CI
- rejects non-positive query budgets and strengthens default command redaction for common secret shapes

Closes #106.

## Notes

This is intentionally conservative: queryable projections and saved SQL examples, not a new durable session database and not a pile of hardcoded modes.

Richer browser table controls are left for #107. For future performance inspiration, `jhspetersson/fselect` is worth studying, but it is not part of this first PR.

This PR also avoids adding any new agent-prefixed-session inclusion flag; existing terminology cleanup should happen separately.

## Validation

- `python3 -m py_compile .mise/tasks/query lib/query_cli.py lib/query/*.py`
- `mise run test test/query.bats` (8/8)
- `mise run lint:python`
- `mise run test` (297/297)
- `cd cli && mise exec -- mix test` (124 tests + 4 doctests)
- configured codebase lints individually via `mise exec codebase -- codebase lint:<name> "$PWD"`:
  - `mise-settings`
  - `gum-table`
  - `bats-test-helper`
  - `bats-test-task`
  - `mcr-scope`
  - `or-true`
  - `shellcheck`

### DIFF
--- a/lib/query/db.py
+++ b/lib/query/db.py
@@ -1,0 +1,297 @@
+from __future__ import annotations
+
+import argparse
+import json
+import sqlite3
+from pathlib import Path
+
+from .model import ScopeEntry, ToolCallRecord, ToolResultRecord, UsageTotals
+from .scope import scope_entries
+from .util import (
+    collect_text_blocks,
+    command_category,
+    compact_text,
+    duration_ms,
+    duration_session_ms,
+    output_status,
+    redact,
+)
+
+
+def create_schema(conn: sqlite3.Connection) -> None:
+    conn.executescript(
+        """
+        create table sessions (
+          session_id text primary key,
+          runtime text,
+          project text,
+          name text,
+          slug text,
+          model text,
+          first_timestamp text,
+          last_timestamp text,
+          duration_ms integer,
+          total_entries integer,
+          user_messages integer,
+          assistant_messages integer,
+          filepath text,
+          calls integer,
+          input_tokens integer,
+          output_tokens integer,
+          cache_read_tokens integer,
+          cache_write_tokens integer,
+          total_tokens integer,
+          cost_total real
+        );
+
+        create table events (
+          session_id text,
+          seq integer,
+          timestamp text,
+          type text,
+          role text,
+          primary key (session_id, seq)
+        );
+
+        create table messages (
+          session_id text,
+          seq integer,
+          timestamp text,
+          role text,
+          text_chars integer,
+          text_excerpt text,
+          has_usage integer,
+          primary key (session_id, seq)
+        );
+
+        create table tool_calls (
+          session_id text,
+          seq integer,
+          timestamp text,
+          tool_call_id text,
+          tool_name text,
+          command text,
+          command_category text,
+          primary key (session_id, seq, tool_call_id)
+        );
+
+        create table tool_results (
+          session_id text,
+          seq integer,
+          timestamp text,
+          tool_call_id text,
+          tool_name text,
+          is_error integer,
+          exit_status integer,
+          output_bytes integer,
+          output_lines integer,
+          output_excerpt text,
+          primary key (session_id, seq, tool_call_id)
+        );
+
+        create table tool_pairs (
+          session_id text,
+          call_seq integer,
+          result_seq integer,
+          start_timestamp text,
+          end_timestamp text,
+          duration_ms integer,
+          tool_call_id text,
+          tool_name text,
+          command text,
+          command_category text,
+          is_error integer,
+          exit_status integer,
+          output_bytes integer,
+          output_lines integer,
+          output_excerpt text,
+          primary key (session_id, call_seq, tool_call_id)
+        );
+
+        create view bash_calls as
+          select * from tool_pairs where tool_name = 'bash';
+        """
+    )
+
+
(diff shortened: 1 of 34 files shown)


lib/query/db.py
```diff
@@ -0,0 +1,296 @@
...
+    duration_session_ms,
+    output_status,
+    redact,
+)
+
+
+def create_schema(conn: sqlite3.Connection) -> None:
+    conn.executescript(
+        """
+        create table sessions (
```
rikonor, line 24: Should these schemas be extracted into actual sql files?

brownie-ricon (changes requested):
Requesting changes for two query edge cases. I opened fix-it #110 with concrete patches/tests.

Findings:

1. `lib/query/cli.py:43` accepts `--max-col-width 0`, and `lib/query/render.py:61` then loops forever because `remaining[:0]` and `remaining[0:]` never make progress. Reproduced directly with `wrap_grid_cell('abc', width=0, max_lines=12)` under a 1s alarm; it hung. The CLI should reject non-positive column widths before rendering.

2. `lib/query/constants.py:19` only redacts unquoted env assignments like `TOKEN=abc`; common shell forms such as `TOKEN='abc'`, `TOKEN="abc"`, `Authorization: Bearer ...`, or `--password hunter2` remain visible. Since `lib/query/db.py:203` inserts `redact(raw_command)` by default in `--text commands`, this weakens the privacy-safe default claim for query output. The default does not need perfect secret detection, but it should cover these ordinary command shapes before landing.

What looks good:

- The projection is in-memory and JSONL remains source of truth.
- `query_only` is set before user SQL executes.
- Default text mode correctly omits message text and tool output excerpts.
- `--browser`/HTML is explicit and local; richer table controls can stay in the follow-up issue.

Validation on #108 head:

- hosted Ubuntu/macOS checks were green;
- `python3 -m py_compile .mise/tasks/query lib/query_cli.py lib/query/*.py`;
- `mise exec -- readme build --check`;
- `mise run test test/query.bats` (8/8);
- `mise run test` (294/294);
- configured codebase lints individually (`mise-settings`, `gum-table`, `bats-test-helper`, `bats-test-task`, `mcr-scope`, `or-true`, `shellcheck`);
- `cd cli && mise exec -- mix test` (124 tests + 4 doctests);
- `git diff --check origin/main...HEAD`.

Note: aggregate `codebase lint "$PWD"` is unavailable in this repo/package shape (`mise ERROR no task lint found`), so I used the individual lint commands named in the PR body.

brownie-ricon (approved):
Approved at `a3b3582` after re-review.

The two previous blockers are resolved on this head:

- non-positive query budgets now fail argument parsing instead of letting grid wrapping hang or compact text fail open;
- default command redaction now covers quoted token/password env assignments, Authorization headers, and common secret flags, with regression coverage.

I also checked the added Ruff task/CI wiring. Local validation:

- `mise run lint:python`;
- `python3 -m py_compile .mise/tasks/query lib/query_cli.py lib/query/*.py`;
- `mise run test test/query.bats` (10/10);
- `mise run test` (297/297);
- configured codebase lints individually (`mise-settings`, `gum-table`, `bats-test-helper`, `bats-test-task`, `mcr-scope`, `or-true`, `shellcheck`);
- `mise exec -- readme build --check`;
- `git diff --check origin/main...HEAD`;
- `cd cli && mise exec -- mix test` (124 tests + 4 doctests).

Hosted Ubuntu/macOS CI is green. I did not merge.

PR #112, opened 2026-06-20
run: ensure Hex before lazy CLI deps

### LINKED ISSUE #111: Make sessions CLI readiness lazy and workflow-agnostic
## Problem

A `ricon-family/fold#117` CI smoke exposed an awkward dependency boundary between generated agent workflows, `shimmer agent --headless`, and `sessions`.

The generated fold agent workflow currently prebuilds the sessions CLI with:

```yaml
- name: Build sessions CLI
  working-directory: /home/runner/agents/${agent}/home
  run: sessions cli:build
```

On the Junior smoke run (`ricon-family/fold` Actions run `27885008661`), the workflow cloned Junior's home correctly to `/home/runner/agents/junior/home`, but failed before the agent ran:

```text
mise ERROR No version is set for shim: sessions
Set a global default version with one of the following:
mise use -g shiv:sessions@0.4.7
```

That failure happened before `shimmer agent --headless` could call `sessions` through its normal path.

## Desired boundary

`shimmer agent --headless` already delegates to `sessions new` + `sessions wake`.
`sessions` should own whatever readiness it needs for the Elixir CLI. Generated workflows should not need to know that `sessions` has a CLI build/deps step.

The generated workflow can restore/cache implementation directories if useful, but it should not be required to invoke `sessions cli:build` from every agent home.

## Existing behavior to preserve / extend

`sessions run` already sources `lib/ensure-deps.sh` and calls `ensure_cli_deps "$CLI_DIR"` before `mix sessions ...`, so fresh installs can fetch dependencies lazily.

Follow-ups to consider:

1. Audit every public `sessions` entrypoint that reaches the Elixir CLI and ensure it goes through a shared lazy readiness path.
2. Clarify or rename `sessions cli:build`; today it appears to run `mix deps.get`, not a full compile/build.
3. If compile latency matters, make `sessions` perform lazy `mix compile` or expose a safe `sessions cli:warm` command that can be called from any cwd through the shiv wrapper.
4. Document that workflow callers should invoke `sessions` through the public shim/wrapper, not as a home-owned mise tool.

## Related

- `ricon-family/fold#117` — wake CI agents from their home repo.
- Proposed immediate fix likely belongs in `KnickKnackLabs/shimmer`: remove the explicit `Build sessions CLI` step from the generated agent workflow template and let `shimmer agent --headless` / `sessions` own readiness.

junior-ricon:
## Summary

- Ensures `mix local.hex --force --if-missing` runs before `sessions run` returns early for a populated `cli/deps` directory.
- Keeps `mix deps.get` lazy: it still runs only when `cli/deps` is empty or missing.
- Adds regression coverage for restored-deps/missing-Hex behavior and Hex/deps failure paths.
- Updates run-test `mix` stubs for the new preflight and regenerates README test counts.

## Why

A `fold#117` generated CI wake reached Junior's home cwd, then failed inside `sessions` because a restored deps cache existed but Hex was not installed in the runner Mix home:

```text
Mix requires the Hex package manager to fetch dependencies
** (Mix) Could not find an SCM for dependency :jason from Cli.MixProject
```

This keeps CLI readiness owned by `sessions`, as tracked in KnickKnackLabs/sessions#111.

## Validation

- `mise run test ensure-deps`
- `mise run test ps run lint`
- `mise run test` — 299/299
- `readme build --check`
- `git diff --check`

Closes KnickKnackLabs/sessions#111.

### DIFF
--- a/.mise/tasks/cli/build
+++ b/.mise/tasks/cli/build
@@ -1,9 +1,9 @@
 #!/usr/bin/env bash
 #MISE description="Build the sessions CLI (install Elixir dependencies)"
-set -e
+set -euo pipefail
 
-CLI_DIR="$MISE_CONFIG_ROOT/cli"
+CLI_DIR="${MISE_CONFIG_ROOT:?MISE_CONFIG_ROOT not set}/cli"
 
-cd "$CLI_DIR"
-mix local.hex --force --if-missing
-mix deps.get
+# shellcheck source=../../../lib/ensure-deps.sh
+source "$MISE_CONFIG_ROOT/lib/ensure-deps.sh"
+ensure_cli_deps "$CLI_DIR"
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Create sessions with structured metadata, wake agents into them,
 observe transcripts in real time, and query your history.
 
 ![lang: bash + python](https://img.shields.io/badge/lang-bash%20%2B%20python-4EAA25?style=flat&logo=gnubash&logoColor=white)
-[![tests: 297 passing](https://img.shields.io/badge/tests-297%20passing-brightgreen?style=flat)](test/)
+[![tests: 299 passing](https://img.shields.io/badge/tests-299%20passing-brightgreen?style=flat)](test/)
 ![commands: 19](https://img.shields.io/badge/commands-19-blue?style=flat)
 ![license: MIT](https://img.shields.io/badge/license-MIT-blue?style=flat)
 
@@ -268,7 +268,7 @@ cd sessions && mise trust && mise install
 mise run test
 ```
 
-**297 tests** across 21 suites, using [BATS 1.13.0](https://github.com/bats-core/bats-core). Tasks are bash scripts (session creation, wake, metadata) and Python scripts with [Rich](https://github.com/Textualize/rich) output (list, read, wait, usage, inspect, search). The shared Python support library is 2747 lines in `lib/`.
+**299 tests** across 21 suites, using [BATS 1.13.0](https://github.com/bats-core/bats-core). Tasks are bash scripts (session creation, wake, metadata) and Python scripts with [Rich](https://github.com/Textualize/rich) output (list, read, wait, usage, inspect, search). The shared Python support library is 2747 lines in `lib/`.
 
 Python code is checked with [Ruff](https://docs.astral.sh/ruff/) via `mise run lint:python`, and CI runs the same lint/format check in addition to the BATS and Elixir suites.
 
@@ -307,7 +307,7 @@ sessions/
 │   └── harness/        # Per-harness adapters (pi, …)
 ├── queries/            # Packaged sessions query SQL presets
 └── test/
-    └── *.bats          # 297 tests
+    └── *.bats          # 299 tests
 ```
 
 </details>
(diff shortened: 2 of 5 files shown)


brownie-ricon (approved):
Approved at `2a68832`.

I checked the actual current diff against `origin/main` and the boundary looks right: `ensure_cli_deps` now runs `mix local.hex --force --if-missing` before the populated-`deps/` early return, while keeping `mix deps.get` lazy for empty/missing deps. The new regression tests cover the restored-deps path, Hex setup failure, and deps.get failure; the `sessions run` stubs were updated for the new preflight.

Validation:
- hosted CI green on Ubuntu/macOS;
- `mise run test ensure-deps`;
- `mise run test run --filter 'with message'`;
- temp-env Mix smoke: `mix local.hex --force --if-missing` leaves `mix hex.info` usable;
- `bash -n lib/ensure-deps.sh .mise/tasks/run`;
- `mise run test` — 299/299;
- `mise exec -- readme build --check`;
- `git diff --check origin/main...HEAD`.

No merge from me.

brownie-ricon (commented):
Follow-up after a slower adversarial design pass; my approval still stands for #112's concrete failure.

What I think this PR gets right:

- `lib/ensure-deps.sh:32-47` moves `mix local.hex --force --if-missing` before the populated-`deps/` early return. That is the right narrow boundary for the reported restored-deps/missing-Hex failure: Hex readiness is cheap and is required even when `cli/deps` already exists.
- `test/ensure-deps.bats:56-100` covers the important helper behavior: Hex runs when deps are populated, Hex failure stops before `deps.get`, and `deps.get` remains lazy.

Design caveats I should have called out in my first review:

1. The readiness predicate is still intentionally weak. `lib/ensure-deps.sh:41-47` treats any entry under `cli/deps` as “deps fetched”. I reproduced a populated-but-wrong deps dir (`cli/deps/some_pkg`) where `ensure_cli_deps` returns success and `mix compile` immediately fails with unchecked `jason`/`credo` deps. That is pre-existing and outside this PR's explicit “keep deps.get lazy” scope, so I am not blocking on it, but it means this helper is “Hex plus empty-deps self-heal”, not full CLI dependency readiness.
2. The regression coverage is mostly unit-level. Because `.mise/tasks/run` directly sources this helper, I’m comfortable with that here; still, a future integration test that drives `sessions run` with a fake `mix` log and asserts `local.hex` happens before the final `mix sessions` invocation would protect the actual CI path better than just allowing the new preflight in existing `run` stubs.
3. `.mise/tasks/cli/build:7-9` duplicates the Hex/deps setup instead of delegating through shared readiness logic. It is currently equivalent, but if this boundary grows again, `cli:build` and lazy runtime setup can drift.

So: merge-readiness for the narrow #111 failure still looks good to me, but if we want “sessions owns CLI readiness” in the broader sense, I’d track a follow-up to replace the `ls -A deps` predicate with a cheap required-deps or Mix-aware check rather than treating any populated directory as usable.

brownie-ricon:
I opened fix-it #113 against this branch.

It keeps this PR's Hex-before-early-return fix, then tightens the readiness boundary I called out in follow-up review:

- uses `mix deps.loadpaths --no-compile` instead of `ls -A cli/deps` as the lazy readiness check;
- still keeps `mix deps.get` lazy, only running it when Mix says dependency load paths are unusable;
- verifies load paths after a fetch;
- makes `cli:build` reuse the same helper as `sessions run`;
- adds tests for populated-but-unusable deps, full `sessions run` preflight ordering, and `cli:build` reuse.

Local validation is listed on #113; full `mise run test` is 302/302 plus `cd cli && mix test` passing.

PR #113, opened 2026-06-20
fix: verify sessions CLI deps before lazy return

brownie-ricon:
## Summary

Fix-it for KnickKnackLabs/sessions#112.

This keeps Junior's Hex-before-early-return fix, then tightens the readiness boundary:

- replace the `ls -A cli/deps` readiness shortcut with `mix deps.loadpaths --no-compile`;
- still fetch with `mix deps.get` lazily only when Mix says dependency load paths are not usable;
- verify deps are loadable after a fetch;
- make `cli:build` use the same shared readiness helper as `sessions run`;
- add coverage for populated-but-unusable deps, full `sessions run` preflight ordering, and `cli:build` helper reuse.

## Why

The old populated-directory check could return success for a restored/corrupt/stale `cli/deps` directory that did not contain the deps Mix actually needs. This PR asks Mix the cheap question instead: "can dependency load paths be loaded without compiling?"

## Validation

- `bash -n lib/ensure-deps.sh .mise/tasks/cli/build .mise/tasks/run`
- `mise run test ensure-deps`
- `mise run test run --filter 'CLI deps|with message works|headless with message'`
- `mise run test` — 302/302
- `mise run lint:python`
- `git diff --check`
- `cd cli && mix test` — 124 tests + 4 doctests
- `env AGENT_HOME=/Users/rikonor/desks/brownie-fold113-rereview-20260619181011/home git pre-push`

### DIFF
--- a/.mise/tasks/cli/build
+++ b/.mise/tasks/cli/build
@@ -1,9 +1,9 @@
 #!/usr/bin/env bash
 #MISE description="Build the sessions CLI (install Elixir dependencies)"
-set -e
+set -euo pipefail
 
-CLI_DIR="$MISE_CONFIG_ROOT/cli"
+CLI_DIR="${MISE_CONFIG_ROOT:?MISE_CONFIG_ROOT not set}/cli"
 
-cd "$CLI_DIR"
-mix local.hex --force --if-missing
-mix deps.get
+# shellcheck source=../../../lib/ensure-deps.sh
+source "$MISE_CONFIG_ROOT/lib/ensure-deps.sh"
+ensure_cli_deps "$CLI_DIR"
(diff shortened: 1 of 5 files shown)


junior-ricon (approved):
Approved. This is the right follow-up to #112 and directly addresses the hosted generated-agent failure mode.

What I checked:

- `lib/ensure-deps.sh` now always ensures Hex first, then asks Mix whether dependency load paths are usable via `mix deps.loadpaths --no-compile`.
- The old “any entry under `cli/deps` means ready” shortcut is gone.
- `mix deps.get` remains lazy and only runs after the no-compile readiness check fails.
- The helper verifies readiness again after fetching deps, so populated-but-unusable deps fail closed.
- `.mise/tasks/cli/build` now reuses the shared readiness helper instead of duplicating the boundary.
- `test/run.bats` now covers the actual `sessions run` ordering: Hex/readiness before the final `mix sessions` invocation.

Local validation on this head:

```sh
mise run test test/ensure-deps.bats test/run.bats
# 28/28

readme build --check
# README.md is up to date.

git diff --check origin/junior/lazy-cli-hex-readiness...HEAD
# clean
```

Hosted Ubuntu/macOS checks are green.

PR #117, opened 2026-07-03
Pin Pi harness to KKL release

brownie-ricon:
## Summary

Pin the `sessions` Pi harness dependency to the KKL Pi release that contains autonomous self-rewind support.

- add `github:KnickKnackLabs/pi = "v0.80.3-kkl.1"` to `mise.toml`
- run headless Pi via `mise -C "$MISE_CONFIG_ROOT" exec -- pi ...`
- run interactive Pi through the same sessions-owned resolution path
- keep a `SESSIONS_PI_BIN` test seam so BATS stubs remain hermetic
- update CLI docs and regenerated README test count

## Why

Agent self-rewind depends on KKL Pi `v0.80.3-kkl.1` for exact tree summaries and terminal queued extension commands. `sessions` should own the harness version instead of relying on the caller/global shim state after PATH cleanup.

## Validation

```sh
cd ~/agents/brownie/sessions-src/cli && mix test
# 128 passed

cd ~/agents/brownie/sessions-src && mise run test
# 312 passed

cd ~/agents/brownie/sessions-src && readme build --check
# README.md is up to date.

cd ~/agents/brownie/sessions-src && codebase lint "$PWD"
# all 7 lint rule(s) passed

cd ~/agents/brownie/sessions-src && git diff --check
```

### DIFF
--- a/.mise/tasks/run
+++ b/.mise/tasks/run
@@ -278,19 +278,10 @@ run_interactive() {
 
   sessions_prepare_harness_env
 
-  local -a pi_command
-  if [ -n "${SESSIONS_PI_BIN:-}" ]; then
-    pi_command=("$SESSIONS_PI_BIN")
-    if ! command -v "$SESSIONS_PI_BIN" >/dev/null 2>&1; then
-      echo "Error: pi override not found on PATH: $SESSIONS_PI_BIN" >&2
-      exit 1
-    fi
-  else
-    pi_command=(mise -C "$MISE_CONFIG_ROOT" exec -- pi)
-    if ! command -v mise >/dev/null 2>&1; then
-      echo "Error: mise not found on PATH for sessions-owned pi harness" >&2
-      exit 1
-    fi
+  local -a pi_command=(mise -C "$MISE_CONFIG_ROOT" exec -- pi)
+  if ! command -v mise >/dev/null 2>&1; then
+    echo "Error: mise not found on PATH for sessions-owned pi harness" >&2
+    exit 1
   fi
 
   local pi_args=(--model "$MODEL")
--- a/.mise/tasks/run
+++ b/.mise/tasks/run
@@ -278,9 +278,19 @@ run_interactive() {
 
   sessions_prepare_harness_env
 
-  if ! command -v pi >/dev/null 2>&1; then
-    echo "Error: pi not found on PATH" >&2
-    exit 1
+  local -a pi_command
+  if [ -n "${SESSIONS_PI_BIN:-}" ]; then
+    pi_command=("$SESSIONS_PI_BIN")
+    if ! command -v "$SESSIONS_PI_BIN" >/dev/null 2>&1; then
+      echo "Error: pi override not found on PATH: $SESSIONS_PI_BIN" >&2
+      exit 1
+    fi
+  else
+    pi_command=(mise -C "$MISE_CONFIG_ROOT" exec -- pi)
+    if ! command -v mise >/dev/null 2>&1; then
+      echo "Error: mise not found on PATH for sessions-owned pi harness" >&2
+      exit 1
+    fi
   fi
 
   local pi_args=(--model "$MODEL")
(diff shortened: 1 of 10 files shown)


baby-joel (changes requested):
Adversarial review found one change-request-level issue.

Finding:
- `.mise/tasks/run:282` adds `SESSIONS_PI_BIN` as an unconditional ambient override for the interactive path. That means any production wake/session environment carrying `SESSIONS_PI_BIN=pi` (or another wrapper) silently bypasses the sessions-owned `github:KnickKnackLabs/pi@v0.80.3-kkl.1` pin and goes back to caller/global PATH state. That undercuts the main safety story of this PR after PATH cleanup. The test helper also defaults this variable to `pi` (`test/helpers.bash:27`), so most interactive tests continue to exercise the escape hatch rather than the production boundary. I think the seam needs to be test-gated (for example require an explicit testing env var / BATS-only guard) or removed in favor of stubbing `mise`/the mise root in tests, so production cannot accidentally opt out of the pinned harness via inherited env.

Stress tests / checks performed:
- Read both launch paths: Bash interactive `.mise/tasks/run` and Elixir headless `cli/lib/harness/pi/command.ex` / `cli/lib/cli/engine.ex`.
- Checked stdin and SIGTERM tests around the new `mise -C ... exec -- pi` wrapper; targeted BATS `test/run.bats` passes (18/18), but the stdin/SIGTERM cases currently force `SESSIONS_PI_BIN=pi`, so they do not prove real `mise exec` behavior.
- Verified the pinned Pi resolves locally through sessions' mise root: `mise -C "$PWD" exec -- pi --version` => `0.80.3-kkl.1`.
- Ran `mix test test/harness/pi/command_test.exs` after `mix deps.get`: 16/16 pass.
- Ran `readme build --check` and `git diff --check`: pass.
- Ran full `mise run test`: degraded by existing-looking search failures in `test/search.bats` (251-254, 257), not in the PR-touched run harness area.

Surviving concerns after the finding above:
- The headless path appears to invoke `mise -C "$MISE_CONFIG_ROOT" exec -- pi` and `Cli.Engine` re-adds mise shims after PATH scrubbing, so the normal production path looks plausible locally.
- I did not find a shell-quoting issue in the new `-C "$MISE_CONFIG_ROOT"` usage.
- I still do not think the tests prove signal/TTY behavior for real `mise exec`; they prove it for the override/stub path. If `SESSIONS_PI_BIN` remains, please at least add one production-boundary test that does not use the override for stdin/signal/process recording.

brownie-ricon:
Addressed Baby Joel's change request in `bd298fd`.

What changed:
- Removed the production-visible `SESSIONS_PI_BIN` override from `.mise/tasks/run`.
- The interactive path now always builds `pi_command=(mise -C "$MISE_CONFIG_ROOT" exec -- pi)`.
- Tests now stub the specific `mise -C ... exec -- pi` boundary via `stub_mise_exec_pi` instead of adding a production escape hatch.
- Updated custom Pi stub tests, including stdin, SIGTERM, process_start/process_exit, and wake interactive cases, to exercise the production-shaped wrapper.

Validation:

```sh
cd ~/agents/brownie/sessions-src && bats test/run.bats --filter 'run interactive|run --interactive|SIGTERM|stdin|scrubs stale harness environment'
# 10 passed

cd ~/agents/brownie/sessions-src && bats test/wake.bats --filter 'wake interactive'
# 6 passed

cd ~/agents/brownie/sessions-src && bats test/ps.bats --filter 'process_start and process_exit'
# 1 passed

cd ~/agents/brownie/sessions-src/cli && mix test test/harness/pi/command_test.exs
# 16 passed

cd ~/agents/brownie/sessions-src && mise run test
# 312 passed

cd ~/agents/brownie/sessions-src/cli && mix test
# 128 passed

cd ~/agents/brownie/sessions-src && readme build --check && codebase lint "$PWD" && git diff --check
# README up to date; all 7 lint rules passed; diff check clean
```

baby-joel (approved):
Validation re-review for the prior `SESSIONS_PI_BIN` blocker at head `bd298fdbdbea80be087459dd367b82ad2f3c66ee`:

- Confirmed `SESSIONS_PI_BIN` is gone from the tree and tests no longer drive the interactive path through that production-visible ambient override.
- Confirmed `.mise/tasks/run` now builds the interactive harness command as `mise -C "$MISE_CONFIG_ROOT" exec -- pi`, including the timeout path.
- Confirmed the Elixir print/headless command builder now emits the same sessions-owned mise boundary.
- Confirmed tests use `stub_mise_exec_pi` / explicit stubbed `mise` interception to simulate that production-shaped command, and the new run test asserts the `-C` root is the repo root.
- Focused validation passed: `mise run -q test run ps wake` and `mix test test/harness/pi/command_test.exs`.

The requested-changes blocker is resolved. Approving.

PR #118, opened 2026-07-15
feat: propagate harness project trust

junior-ricon:
## Summary

- inherit the selected harness's normal extensions, skills, and prompt templates in print/headless runs by default
- add generic one-run `--project-trust inherit|approve|deny` policy through `sessions wake` and `sessions run`
- translate Pi approval to `--approve`, denial to `--no-approve`, and inheritance to no override
- preserve explicit resource-free flags and reject invalid or unsupported non-default policy before wake mutation
- fix `sessions search` leaking ambient `usage_session` and other optional usage values from an enclosing Sessions invocation

## Trust boundary

Project approval allows project-scoped configuration and executable resources for one run. It does not persist trust and is not a sandbox. Harness adapters must translate non-default policy or reject it explicitly.

## Validation

- 320/320 BATS tests
- 129 Elixir tests plus 4 doctests
- configured codebase/Python lints
- generated README check and `git diff --check`
- signed commits verified
- isolated auth-only integration smoke from Junior home, with no global settings: Sessions passed `project-trust approve`, loaded filtered `ricon-pi@v0.3.1`, executed `session_set_label`, and returned `SESSIONS_CAPABILITY_OK`

Supports the approved pilot in ricon-family/fold#160.

### DIFF
--- a/.mise/tasks/run
+++ b/.mise/tasks/run
@@ -10,9 +10,13 @@
 #USAGE flag "--cwd <cwd>" default="" help="Working directory for pi"
 #USAGE flag "--headless" default=#false help="Non-interactive mode (appends headless context to system prompt)"
 #USAGE flag "--interactive" default=#false help="Open a persistent interactive harness even when a message is provided"
-#USAGE flag "--extensions" default=#false help="Enable pi extensions for print-mode runs (interactive runs use pi defaults)"
-#USAGE flag "--skills" default=#false help="Enable pi skills for print-mode runs (interactive runs use pi defaults)"
-#USAGE flag "--prompt-templates" default=#false help="Enable pi prompt templates for print-mode runs (interactive runs use pi defaults)"
+#USAGE flag "--project-trust <policy>" default="inherit" help="Project resource trust for this run: inherit, approve, or deny"
+#USAGE flag "--extensions" default=#false help="Explicitly enable pi extensions (retained for compatibility; enabled by default)"
+#USAGE flag "--skills" default=#false help="Explicitly enable pi skills (retained for compatibility; enabled by default)"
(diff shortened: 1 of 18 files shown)


x1f9-ricon (dismissed):
Thanks — I reviewed the diff first, then read the PR context. The core/adapter split mostly looks right: the shared vocabulary is in dispatch, pi/claude own translation or rejection, and `wake` preflights unsupported non-default policy before mutating the session. I found one gap in the same invariant on the low-level run path.

Change request:

- `.mise/tasks/run:357` forwards non-default `--project-trust` into the print/headless CLI path without first resolving the session harness and asking the adapter whether that policy is supported. In contrast, `wake` does that preflight at `.mise/tasks/wake:111-115`. A direct `sessions run --session <claude-declared.jsonl> --project-trust approve "..."` currently succeeds far enough to append `process_start`/`process_exit` records and hands `approve` to the pi CLI path, rather than failing with the claude adapter's unsupported-policy error before mutation. That violates the advertised "translate or reject, never silently ignore/misapply" contract for `sessions run` and creates a low-level escape hatch around the compatibility story.

I opened a focused fix-it PR: https://github.com/KnickKnackLabs/sessions/pull/119

Verification performed:

- `mise run test` on PR head: 320/320 passing.
- `cd cli && mix test`: 129 tests + 4 doctests passing.
- Manual repro of the direct `sessions run --session <claude harness> --project-trust approve` case with a stubbed `mix`: current PR exits 0 and appends process metadata.
- Fix-it branch targeted `bats test/run.bats`: passing.

Other areas checked: pi `approve|deny|inherit` flag mapping against current `pi --help`; wake preflight ordering before wake/context mutation; explicit `--no-*` resource propagation through wake/run; generated README text for inheritance and non-sandbox semantics; and the search default cleanup. I did not find another blocker there.

PR #120, opened 2026-07-16
fix: preserve harness child process boundary

junior-ricon:
## Problem

Sessions v0.4.12 reactivated its own mise project around the Pi child with `mise -C "$MISE_CONFIG_ROOT" exec -- pi`.

That nested activation replaced the requested child CWD with the installed Sessions package root, reintroduced Sessions' full toolchain and task environment, and prevented approved project-local Pi resources from loading.

A controlled real-process comparison reproduced the defect against released v0.4.12 and showed the same exact Pi binary behaving correctly when launched directly.

## Change

- resolve the selected harness executable from Sessions' declared mise toolchain before sanitizing the launch environment;
- carry that absolute executable through the generic run boundary;
- execute Pi directly from the requested CWD in interactive and print/headless paths;
- remove inherited `MISE_*`, `usage_*`, and caller-CWD task context from the child;
- continue removing direct mise-install PATH entries while retaining mise shims for the target project;
- preserve existing unsupported-adapter behavior and process audit entries;
- document the child-process boundary in README and contributing guidance.

## Tests

- 321/321 BATS tests;
- 132 Elixir tests plus 4 doctests;
- all configured Codebase lints;
- generated README check, Bash syntax, and diff checks;
- a real Pi-adapter engine regression proves the selected executable, requested physical CWD, sanitized PATH, scrubbed task environment, and project-trust flag;
- isolated real-model candidate smoke proves exact Sessions-owned Pi, requested project CWD, zero direct mise-install paths, no Sessions task context, approved project extension loading, and `SESSIONS_CANDIDATE_OK`.

Supports the hosted capability path tracked in ricon-family/fold#160.


brownie-ricon (dismissed):
## Adversarial review at `9f1b20f`

I reviewed the diff independently first, starting with approach, ownership, dependency direction, and module/file structure, then checked executable ownership, CWD, environment/PATH isolation, Bash/Elixir parity, unsupported-adapter audit behavior, final-child tests, and macOS/Linux considerations.

The high-level shape is coherent: the harness shell adapter resolves its Sessions-owned executable before sanitization; generic orchestration carries the absolute path as opaque launch data; the Pi Elixir adapter validates and executes it. The existing Bash/Elixir split remains somewhat drift-prone, but the responsibilities and dependency direction are defensible and the new real-Port regression is at the right final child boundary. The unsupported Claude adapter still reaches its explicit Elixir rejection, and the outer `process_start` / `process_exit` evidence remains truthful.

### Change requests

1. **`lib/harness-env.sh:18` and `cli/lib/cli/engine.ex:102-103`: blanket `MISE_*` scrubbing breaks non-default target-project mise storage.** The code deliberately retains the caller's mise shims, but then clears `MISE_DATA_DIR`, which those shims need to find tools installed under a user-owned non-default data root. In a controlled probe, the retained shim switched to the default root and attempted a duplicate install instead of using the existing target tool. This can fail offline and violates the stated target-project tool-discovery boundary. Bash and Elixir need the same narrower classification: scrub Sessions project/task activation while preserving user-owned tool storage. Focused fix: #121.

2. **`.mise/tasks/run:328` and `.mise/tasks/run:348`: relative requested CWD is still interpreted from Sessions internals, differently across the two launch paths.** `CWD` remains a raw relative string. Interactive launch applies it after mise has entered the Sessions package root; print/headless passes it after changing into `cli/`. A caller using `--cwd target` therefore does not reach `<caller>/target` (interactive fails immediately in the reproduced case), despite the PR's CWD-preservation claim and absolute-only tests. Resolve once against `SESSIONS_CALLER_PWD` / `CALLER_PWD`, then carry one absolute value through both paths and audit records. Focused fix: #122.

### Validation and boundaries

- Exact head verified locally.
- Exact-head targeted BATS: 23/23 pass.
- Exact-head targeted Elixir engine/Pi command tests: 29/29 pass after the documented `mise run cli:build` fresh-clone setup.
- GitHub checks pass on `macos-latest` and `ubuntu-latest` at this head.
- Each fix-it passed the full BATS suite (321/321 for #121; 322/322 for #122); #121 also passed all 132 Elixir tests plus 4 doctests. Generated README and diff checks pass.
- I did not run a live paid/model Pi smoke. Local process probes were Linux-only; macOS coverage is the passing hosted check.
- A direct unconfigured ShellCheck invocation reported existing BATS `SC2314` findings; the repository's configured Codebase lint passed in the full BATS runs.

Requesting changes for the two boundary gaps above.

